PR #1, opened 2020-11-16
Set up build tools for the registry

johnmcollier:
Fixes: https://github.com/devfile/api/issues/164

This PR does the following:

1) Copies over the devfiles from [odo-registry/devfiles](https://github.com/odo-devfiles/registry) and removes the sample devfiles that were in this repository
    - Devfiles were copied as-is

2) Adds scripts for building the index.json and building it into an image
    - `build.sh` uses the index-generator tool to generate the index.json and builds a Docker container image containing the stacks and index.json
      - Any validation of the devfile stacks will be done as part of `index-generator`, once that tool has been updated accordingly

3) Adds GitHub workflow actions
    - Action for running the build + validation on PRs
    - Action for pushing new images up to quay.io/devfile/devfile-index:next on each merge to `master`.


build.sh
```diff
@@ -0,0 +1,51 @@
...
+# Build the index generator/validator
+echo "Building index-generator tool"
+cd ./registry-support/index/generator
+./build.sh
+if [ ! $? -eq 0 ]; then
+  echo "Failed to build index-generator tool"
+  exit 1
+fi
+echo -e "Successfully built the index-generator tool\n"
+cd "$OLDPWD"
```
GeekArthur, line 32: Seems like we should define the `OLDPWD` before building the index generator?

index.json
```diff
@@ -0,0 +1,112 @@
+[
```
GeekArthur, line 1: Given on the image building stage we use the index generator to generate the index.json on the fly, I am wondering do we still need the index.json file in this repo?

GeekArthur (changes requested):

index.json
johnmcollier: Oops, that definitely was not supposed to get committed. Will remove 😅

Dockerfile
```diff
@@ -0,0 +1,12 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+FROM quay.io/devfile/metadata-server:next
```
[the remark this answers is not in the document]
johnmcollier: Yup, I'll do that as part of my changes in https://github.com/devfile/registry-support/pull/11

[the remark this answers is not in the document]
johnmcollier: @GeekArthur, how about we rename it to something like `devfile-index-base`? To be clear that it's a base image?

build.sh
johnmcollier: So, it actually is a predefined shell environment variable: 
Screen Shot 2020-11-19 at 9 50 50 AM

I was trying to avoid using `cd -`, as it always prints the old directory, and found `$OLDPWD` as an alternative.

Dockerfile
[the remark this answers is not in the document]
GeekArthur: Yeah, `index-server-base` looks good to me.

build.sh
GeekArthur: Oh, Today I Learned!

index.json
GeekArthur: 👍

GeekArthur (approved):
LGTM, thanks for making the changes!

PR #4, opened 2021-01-22
Update registry repo CI to use proper build script

johnmcollier:
Looks like when we moved the build script from this repo to registry-support/build-tools, we never updated the CI in this repo.


GeekArthur (commented):
The PR looks good to me, just one more question, if we switch to use build-tools to build the devfile-index image, should we delete the `build.sh` and `push.sh` scripts in this repo as those are old scripts and we won't use them any more?

johnmcollier:
@GeekArthur D'oh, those should've been deleted 🤦‍♂️

johnmcollier:
Updated!

GeekArthur (approved):
Thanks for making the changes, LGTM!

PR #6, opened 2021-02-10
Merge meta.yaml

GeekArthur:
Signed-off-by: jingfu wang <jingfwan@redhat.com>

This PR merges meta.yaml to devfile.yaml

Related issue: https://github.com/devfile/api/issues/222

### DIFF
--- a/stacks/java-quarkus/devfile.yaml
+++ b/stacks/java-quarkus/devfile.yaml
@@ -2,7 +2,11 @@ schemaVersion: 2.0.0
 metadata:
   name: java-quarkus
   version: 1.1.0
-  website: https://quarkus.io
+  displayName: Quarkus Java
+  description: Upstream Quarkus with Java+GraalVM
+  tags: ["Java", "Quarkus"]
+  projectType: "quarkus"
+  language: "java"
 starterProjects:
   - name: quarkus-ex
     git:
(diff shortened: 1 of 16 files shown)


GeekArthur:
This PR should be merged after this PR https://github.com/devfile/registry-support/pull/28 is merged

stacks/java-quarkus/devfile.yaml
```diff
@@ -2,7 +2,11 @@ schemaVersion: 2.0.0
 metadata:
   name: java-quarkus
   version: 1.1.0
-  website: https://quarkus.io
```
johnmcollier, line 5: Do we want to remove the project field from the devfile?

johnmcollier (commented):
Left one small comment, otherwise looks good

stacks/java-quarkus/devfile.yaml
GeekArthur: I checked the metadata field of API reference: https://docs.devfile.io/devfile/2.1.0/user-guide/api-reference.html and found the website field is not defined under metadata field, that's why I removed that, but I am open to add it back to both this devfile.yaml and API reference as I am gonna create another PR for API reference to add some fields (eg. projectType, language) under metadata.

GeekArthur: Do you think we should add the website field back to both this devfile.yaml and API reference? Currently in our registry repo, only java-quarkus has the website field under metadata.

johnmcollier (approved):

johnmcollier (approved):

PR #8, opened 2021-02-22
Set environment variables for pushing index image to quay

johnmcollier:
This PR updates the build_and_deploy.sh script that runs on the app-sre jenkins CI job to properly set the `IMAGE` and `IMAGE_TAG` variables. It also makes sure now to push to both to the given tag (corresponding to the git revision) as well as the `next` tag.


GeekArthur (approved):
LGTM

GeekArthur (approved):
LGTM

PR #9, opened 2021-03-02
Update devfiles from odo-devfiles/registry

johnmcollier:
Updates the devfiles in this registry from odo-devfiles/registry


GeekArthur (approved):
LGTM!

PR #10, opened 2021-03-08
Updata java-openliberty from odo-devfiles

johnmcollier:
The java-openliberty devfile was updated on the odo-devfiles/registry repo shortly after the last registry sync that I did, so I'm just updating it here.


GeekArthur (approved):
LGTM

PR #13, opened 2021-03-23
Update devfiles based on odo-devfiles/registry

johnmcollier:
Updates java-quarkus and java-vertx devfiles based on recent changes in odo-devfiles/registry


GeekArthur (approved):
LGTM

PR #15, opened 2021-04-06
Update the Open Liberty devfile to version 0.4.3

mezarin:
Porting PR: https://github.com/odo-devfiles/registry/pull/77
FYI @ajm01, @scottkurz, @elsony

### DIFF
--- a/stacks/java-openliberty/devfile.yaml
+++ b/stacks/java-openliberty/devfile.yaml
@@ -1,13 +1,13 @@
 schemaVersion: 2.0.0
 metadata:
   name: java-openliberty
-  version: 0.4.1
+  version: 0.4.3
   description: Java application stack using Open Liberty runtime
   displayName: "Open Liberty"
   language: "java"
   projectType: "docker"
-  alpha.build-dockerfile: "https://github.com/OpenLiberty/application-stack/releases/download/outer-loop-0.5.0/Dockerfile"
-  alpha.deployment-manifest: "https://github.com/OpenLiberty/application-stack/releases/download/outer-loop-0.5.0/app-deploy.yaml"
+  alpha.build-dockerfile: "https://github.com/OpenLiberty/application-stack/releases/download/outer-loop-0.5.1/Dockerfile"
+  alpha.deployment-manifest: "https://github.com/OpenLiberty/application-stack/releases/download/outer-loop-0.5.1/app-deploy.yaml"
 starterProjects:
   - name: user-app
     git:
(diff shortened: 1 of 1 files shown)


scottkurz:
LGTM

johnmcollier (approved):
LGTM.

elsony (changes requested):
Is there a recent change on the starterProjects: https://github.com/OpenLiberty/application-stack-starters?
I just tried this out and the starter project does nothing (it used to serve the starter page) and the default URL gives a 404 error. It may not be a problem directly associated with this PR. However, I think we should fix the existing starter project problem before putting any more changes to the stack.

scottkurz:
> 
> 
> Is there a recent change on the starterProjects: https://github.com/OpenLiberty/application-stack-starters?
> I just tried this out and the starter project does nothing (it used to serve the starter page) and the default URL gives a 404 error. It may not be a problem directly associated with this PR. However, I think we should fix the existing starter project problem before putting any more changes to the stack.

@elsony  no I don't think we've changed the starter in this regards recently.   It's "working as designed", and no, there's no feedback from the browser that you've created a successful project currently.

We might reconsider that choice, and add something extra for "feedback", but for now this is intentional.

Let me ask a question, though.  In the move from https://github.com/odo-devfiles/registry/ to the current repo, we no longer perform the same validation we did in the Travis build.

Is this performed somewhere else ?  Should we be still doing PRs against the odo devfile registry repo as well?

Any suggestions?

johnmcollier:
> Let me ask a question, though. In the move from https://github.com/odo-devfiles/registry/ to the current repo, we no longer perform the same validation we did in the Travis build.

This repo will be performing that validation / tests as well once https://github.com/openshift/odo/pull/4525 is merged and we have a binary of odo that supports the new registry format

>  Should we be still doing PRs against the odo devfile registry repo as well?

Until odo, and the stack creators (such as yourself and @mezarin) have migrated to the new registry, you should still be sending PRs to odo-devfiles/registry as well. We appreciate you also sending them here too 😄, until the migration is complete.

elsony:
@scottkurz The starter project used to show this when running the application:
image

IMHO, that serves better as a starter project since the user can click on the first link to execute some java code. I think even a simple java program that runs a hello world message is better than resulting in a 404 error by default. If I am a user and I created a start project in odo and run it and the URL returns a 404 to try out the stack. I would get a false impression that the OpenLiberty stack is not working at all (or at least the starter project is not).

scottkurz:
> @scottkurz The starter project used to show this when running the application:
> IMHO, that serves better as a starter project since the user can click on the first link to execute some java code. We may want to do a minor tweak on the message to remove the odo reference since other tools like Che will be consuming the content from the new community registry. I think even a simple java program that runs a hello world message is better than resulting in a 404 error by default. If I am a user and I created a start project in odo to try out the stack and run it but the URL returns a 404. I would get a false impression that the OpenLiberty stack is not working at all (or at least the starter project is not).

@elsony I see your point but it's been like this for some months now and this doesn't have anything to do with this PR with this devfile update. 
I'm thinking of making the change though so we don't have to explain to each tool in the devfile ecosystem that we chose to implement the starter this way.

But unless this is affecting some automated testing (like the odo registry travis build), it seems like a separate issue.  We can change the starter completely independently, the devfile just points to the default branch head.

elsony (approved):
/lgtm

johnmcollier:
@scottkurz @mezarin Heads up that per our update schedule for the regisry, these changes will be reflected in https://registry.devfile.io tomorrow (Wednesday)

PR #17, opened 2021-04-16
Fixing WildFly issue while waiting for fix in WildFly 23.0.1 image

ehsavoie:
* Updating the QS version
* Fixing the WildFly image version to be compatible with the QS


elsony (approved):
/lgtm

I notice this one only contain the fix for java-wildfly but no fix on java-wildfly-bootable-jar fix (as in https://github.com/odo-devfiles/registry/pull/76). Are we missing the bootable jar fix on this one? I'll merge this one anyway. Feel free to raise a new PR for the bootable jar fix where applicable.

PR #18, opened 2021-04-21
Add extraDevfileEntires.yaml to support sample

GeekArthur:
This PR adds extraDevfileEntires.yaml to support sample according to the proposal: https://docs.google.com/document/d/1tbzZK9aWtpw5AtoZmUl1Jy8x8w9N5hhjXu3c26VYiYY/edit#heading=h.6w072oor8mro

Signed-off-by: jingfu wang <jingfwan@redhat.com>

### DIFF
--- a/extraDevfileEntries.yaml
+++ b/extraDevfileEntries.yaml
@@ -4,7 +4,7 @@ samples:
     displayName: Basic NodeJS
     description: A simple Hello World application
     icon: nodejsIcon.svg
-    tags: ["NodeJS", “Express”]
+    tags: ["NodeJS", "Express"]
     git:
       remotes:
         origin: https://github.com/redhat-developer/devfile-sample
--- a/extraDevfileEntries.yaml
+++ b/extraDevfileEntries.yaml
@@ -1,0 +1,18 @@
+schemaVersion: 1.0.0
+samples:
+  - name: nodejs-basic
+    displayName: Basic NodeJS
+    description: A simple Hello World application
+    icon: nodejsIcon.svg
+    tags: ["NodeJS", “Express”]
+    git:
+      remotes:
+        origin: https://github.com/redhat-developer/devfile-sample
+  - name: vertx-secured-http
+    displayName: Vert.x with secured HTTP
+    description: A Vert.x application using secured HTTP
+    icon: vertxIcon.svg
+    tags: ["Java", "Vert.x"]
+    git:
+      remotes:
+        origin: https://github.com/openshift-vertx-examples/vertx-secured-http-example-redhat
--- a/extraDevfileEntries.yaml
+++ b/extraDevfileEntries.yaml
@@ -5,6 +5,8 @@ samples:
     description: A simple Hello World application
     icon: nodejsIcon.svg
     tags: ["NodeJS", "Express"]
+    projectType: nodejs
+    language: nodejs
     git:
       remotes:
         origin: https://github.com/redhat-developer/devfile-sample
@@ -13,6 +15,8 @@ samples:
     description: A Vert.x application using secured HTTP
     icon: vertxIcon.svg
     tags: ["Java", "Vert.x"]
+    projectType: vertx
+    language: java
     git:
       remotes:
         origin: https://github.com/openshift-vertx-examples/vertx-secured-http-example-redhat


extraDevfileEntries.yaml
```diff
@@ -0,0 +1,18 @@
+schemaVersion: 1.0.0
+samples:
+  - name: nodejs-basic
+    displayName: Basic NodeJS
+    description: A simple Hello World application
+    icon: nodejsIcon.svg
+    tags: ["NodeJS", “Express”]
```
elsony: Let's add `language` and `projectType` to the list (to match the stack list):
For `nodejs-basic`,
```
  projectType: nodejs,
  language: nodejs,
```

For vertx-secured-http,
```
  projectType: vertx,
  language: java, 
```

elsony (changes requested):

extraDevfileEntries.yaml
GeekArthur: Good suggestion, will update.

extraDevfileEntries.yaml
```diff
@@ -0,0 +1,18 @@
+schemaVersion: 1.0.0
+samples:
```
maysunfaisal, line 2: @elsony should samples be the top level or a type(type can be stack or sample)?

Proposal 2 has it in the top level and the index output has it as type https://docs.google.com/document/d/1tbzZK9aWtpw5AtoZmUl1Jy8x8w9N5hhjXu3c26VYiYY/edit#heading=h.6w072oor8mro

johnmcollier (commented):
Should we define the schema for each sample "entry" somewhere other than Google Docs?

extraDevfileEntries.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - name: nodejs-basic
```
johnmcollier, line 3: Does each sample need a `version` like we do with stacks?

johnmcollier (changes requested):

extraDevfileEntries.yaml
johnmcollier: Thinking it over further, I think the answer is no, so gonna resolve this review thread.

GeekArthur:
> Can we define the schema for each sample "entry" somewhere other than Google Docs?

Probably we can define it under the [registry structure](https://github.com/devfile/api/blob/main/docs/proposals/registry/registry-structure.md) as extraDevfileEntires.yaml is under the root of the registry, but for future it would be better we have our own registry website (similar to devfile schema) then we can host all the docs and schemas related registry there.

extraDevfileEntries.yaml
GeekArthur: Done

extraDevfileEntries.yaml
elsony: @maysunfaisal When I wrote the proposal, I started with having both samples and stacks under the same list in the `extraDevfileEntries.yaml`. However, I find that I need to add an extra non-functional parent, e.g. `devfiles:` at the top level. I find in that file, it is cleaner to just separate out those two lists. 

Note that I keep `type` field in index.json since the registry just contains a list of devfile entires. Having them in the same list will be naturally having one single list to contain all devfiles in the query in the returned json object.

Do your think it works better to have the type instead of two separate lists in that file? I don't feel strongly about both approaches so I am open to suggestions.

elsony (approved):
approve on my side (under the condition that the comment thread https://github.com/devfile/registry/pull/18/files#r617824628 will be resolved).

elsony:
> > Can we define the schema for each sample "entry" somewhere other than Google Docs?
> 
> Probably we can define it under the [registry structure](https://github.com/devfile/api/blob/main/docs/proposals/registry/registry-structure.md) as extraDevfileEntires.yaml is under the root of the registry, but for future it would be better we have our own registry website (similar to devfile schema) then we can host all the docs and schemas related to registry there.

This PR is just introducing the file itself. For the schema of this file, I assume we'll have one when we do that we do the implementation of consuming this file (that's why I have schema version in the file itself). The way that I see it is that we'll have a branch of docs on the registry from the main landing page (not necessarily to be under a separate website). We can present that branch of the doc as a branch off for different persona, e.g. registry admin. But I agree all registry-related doc + schema (including index.json + extraDevfileEntries.yaml) should go to the same place.

extraDevfileEntries.yaml
elsony: +1 for no need to have a version for each sample entry. Samples usage are supposed to be one-off clone and detach from the origin so I don't think we need to track the version of that. Same reason why we don't have a version on the starterProject entry.

extraDevfileEntries.yaml
maysunfaisal: two lists is fine

extraDevfileEntries.yaml
GeekArthur: @johnmcollier Can you approve this PR if it's resolved and it looks good to you?

johnmcollier (approved):

PR #20, opened 2021-04-23
Add GitHub action for validating devfile stacks

johnmcollier:
This PR ports over the [odo-devfiles/registry test script](https://github.com/odo-devfiles/registry/blob/master/scripts/test.sh) to run on Minikube and as a GitHub action in this repository.
    
   - The test script uses odo to verify that the following commands work for each devfile stack: `odo create`, `odo url create` and `odo push`, and that the starter application is properly exposed (by checking for an HTTP 200 status code)
   - Some small modifications were made to support Minikube instead of Minishift, but otherwise it's unchanged

I also included a brief readme in the `tests/` folder that provides instructions on running the test script locally on Minikube.


tests/test.sh
```diff
@@ -0,0 +1,151 @@
...
+        FAILED_TESTS="$FAILED_TESTS $devfileName"
+        return 1
+    fi
+
+    if [ "$ENV" = "minikube" ]; then
+        # ToDo: Clean up, I'm not happy about having specific checks for the stacks with multiple ports
+        # But since we're testing against minikube, we need to specifically create the URL/ingress before pushing
+        # And if there's multiple ports in the devfile, a port must be specified.
+        if [ "$devfileName" = "java-wildfly" ] || [ "$devfileName" = "java-wildfly-bootable-jar" ]; then
+            $ODO_PATH url create --host "$(minikube ip).nip.io" --port 8080 || error=true
```
maysunfaisal, line 70: another alternative is to create a odo url with all the endpoints and checking one of them returns 200 🤔

maysunfaisal:
@johnmcollier not saying we should do this but another alternative is to perhaps use https://github.com/devfile/api/tree/main/pkg/validation validations in this registry and not relying on devfile consumer to validate 🤔  since registry-support also does this https://github.com/devfile/registry-support/blob/master/index/generator/library/library.go#L109-L113. But odo calls the same func anyways.

tests/test.sh
johnmcollier: Let me see if that can be quickly adapted, if so I’ll add it.

johnmcollier:
@maysunfaisal good suggestion. As part of the registry build, the index generator calls the devfile validation I believe for each devfile (CC @GeekArthur)

tests/test.sh
johnmcollier: @maysunfaisal Updated.

GeekArthur:
> @maysunfaisal good suggestion. As part of the registry build, the index generator calls the devfile validation I believe for each devfile (CC @GeekArthur)

That's correct.

tests/test.sh
```diff
@@ -0,0 +1,164 @@
...
+fi
+if [ "$REGISTRY" != "local" ] && ["$REGISTRY" != "remote" ]; then
+  echo "ERROR: Allowed values for REGISTRY are either \"local\" (default) or \"remote\"."
+  exit 1
+fi
+
+for devfile_dir in $(find $DEVFILES_DIR -maxdepth 1 -type d ! -path $DEVFILES_DIR); do
+    devfile_name="$(basename $devfile_dir)"
+    devfile_path=$devfile_dir/devfile.yaml
+    if [ $devfile_name != "java-wildfly-bootable-jar" ]; then
```
GeekArthur, line 149: I think it would be better to add a comment to explain the reason why we skip `java-wildfly-bootable-jar`?

tests/test.sh
```diff
@@ -0,0 +1,164 @@
...
+
+    # remember if there was en error
+    error=false
+
+    tmpDir=$(mktemp -d)
+    cd "$tmpDir" || return 1
+
+    $ODO_PATH project create "$devfileName" || error=true
+    if $error; then
+        echo "ERROR project create failed"
```
GeekArthur, line 52: Should we add the cleanup step after we hit the error? Probably we can create a clean up function and call it either the test fails in the middle or the test succeeds in the end.

tests/test.sh
```diff
@@ -0,0 +1,164 @@
...
+        $ODO_PATH project delete -f "$devfileName"
+        FAILED_TESTS="$FAILED_TESTS $devfileName"
+        return 1
+    fi
+
+    if [ "$ENV" = "minikube" ]; then
+        # ToDo: Clean up, I'm not happy about having specific checks for the stacks with multiple ports
+        # But since we're testing against minikube, we need to specifically create the URL/ingress before pushing
+        # And if there's multiple ports in the devfile, a port must be specified.
+        if [ "$devfileName" = "java-wildfly" ] || [ "$devfileName" = "java-wildfly-bootable-jar" ]; then
```
GeekArthur, line 74: IIRC, we already skipped `java-wildfly-bootable-jar` when we call the test() ?

tests/test.sh
```diff
@@ -0,0 +1,164 @@
...
+if [ "$REGISTRY" != "local" ] && ["$REGISTRY" != "remote" ]; then
+  echo "ERROR: Allowed values for REGISTRY are either \"local\" (default) or \"remote\"."
+  exit 1
+fi
+
+for devfile_dir in $(find $DEVFILES_DIR -maxdepth 1 -type d ! -path $DEVFILES_DIR); do
+    devfile_name="$(basename $devfile_dir)"
+    devfile_path=$devfile_dir/devfile.yaml
+    if [ $devfile_name != "java-wildfly-bootable-jar" ]; then
+      test "$devfile_name" "$devfile_path"
```
GeekArthur, line 150: I think we need to check the return code of test() here. Then if the test() fails we can either exit the test script immediately or record the failure and return all the results at the end of the test script.

tests/test.sh
```diff
@@ -0,0 +1,164 @@
...
+    fi
+done
+
+# remember if there was an error so the script can exist with proper exit code at the end
+error=false
+
+# print out which tests failed
+if [ "$FAILED_TESTS" != "" ]; then
```
GeekArthur, line 158: I wonder if we only catch the last failed stack test as FAILED_TESTS can be override each time we can test().

GeekArthur (changes requested):

tests/test.sh
johnmcollier: We need to leave it in since the java-wildfly-bootable-jar stack will be fixed soon.

tests/test.sh
johnmcollier: It works as expected:
```
+ echo 'FAILURE: FAILED TESTS:  java-openliberty java-wildfly-bootable-jar'
FAILURE: FAILED TESTS:  java-openliberty java-wildfly-bootable-jar
```

tests/test.sh
johnmcollier: Sure, that's reasonable.

tests/test.sh
johnmcollier: I'd say no, it's in a temp folder that will be removed upon reboot/restart.

tests/test.sh
johnmcollier: I'd say let's leave it as-is for now, since it works fine either way. We can always revisit later if needed.

tests/test.sh
johnmcollier: Updated.

GeekArthur (approved):

PR #22, opened 2021-04-27
Fix GitHub actions for updated build tools

johnmcollier:
We should be calling `build_image.sh` in the Github actions for this repo now rather than `build.sh`


GeekArthur (approved):
LGTM

GeekArthur (approved):
LGTM

PR #25, opened 2021-05-04
Add Dockerfile and OpenShift CI script

johnmcollier:
- Adds a Dockerfile for building the registry index container image on the OpenShift CI
- Adds an `openshift_integration.sh` test script to be run on the OpenShift CI. That does the following:
    - Downloads and builds odo from master
    - Deploys the OpenShift templates for the devfile registry, using the devfile registry index image passed in `REGISTRY_IMAGE`
    - Runs the test scripts against the devfile registry that was deployed.


johnmcollier:
@josephca Please review

PR #33, opened 2021-06-02
Update sample icon link

GeekArthur:
Given we only store the sample metadata in the registry instead of storing all the sample resource, then in the icon field of the sample metadata we should use the full icon URL instead of file path as we don't store the icon in our registry.

Signed-off-by: Jingfu Wang <jingfwan@redhat.com>


maysunfaisal (approved):
sample icons can be full URIs too

PR #42, opened 2021-06-23
Update dockerfile to use yq

johnmcollier:
The updated build tools now require yq, so updated the dockerfile to add it. 

The index server base image now expects the stacks and samples to be under `/registry` so updating the dockerfile for that too.


GeekArthur (approved):
LGTM

PR #45, opened 2021-07-08
Add pull request template

johnmcollier:
Adds Pull Request template for the registry repository. Also updates the contributing guide a bit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ Updating an existing devfile stack is relatively straightforward:
 The devfile samples used in this devfile registry are stored in the `extraDevfileEntries.yaml` file in the root of the repository. To add a devfile sample:
 
 1) Open `extraDevfileEntries.yaml` in an editor
-2) Add an entry to the file similar to the following:
+2) Add an entry to the file with the following required fields:
 ```
   - name: <sample-name>
     displayName: <sample-display-name>
@@ -73,7 +73,7 @@ The devfile samples used in this devfile registry are stored in the `extraDevfil
       remotes:
         origin: <link-to-sample-git-repository>
 ```
-3) Fill in the fields in the angle brackets based on your sample.
+3) Fill in the fields in the angle brackets based on your sample. Note that there must be only one git remote for the devfile sample. 
 4) Open a pull request against this repository with your changes.
 
 ### Updating
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,6 @@ Updating an existing devfile stack is relatively straightforward:
 4) Test your changes:
     
     - Minimally, testing with odo (`odo create`, `odo push`, etc) is recommended, however if your Devfile is used with other tools, it's recommended to test there as well
-    - The 
 5) Open a pull request against this repository with your changes
 
 ## Samples
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,7 @@ The following are required to build the devfile index container image containing
 Updating an existing devfile stack is relatively straightforward:
 
 1) Find the stack under the `stacks/` folder that you wish to update.
-2) Make the necessary changes to the stack (such as updating image tags, commands, starter projects, etc).
+2) Make the necessary changes to the stack, such as: updating image tags, commands, starter projects, etc.
 3) Update the version of stack, following the [semantic versioning format](https://semver.org/).
 4) Test your changes:
     
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing to this registry
 
-This document outlines the requirements for contributing a devfile stack or sample to this repository
+This document outlines the requirements for contributing a devfile stack or sample to this repository.
 
 The [devfile registry structure](https://github.com/devfile/api/blob/main/docs/proposals/registry/registry-structure.md#repository-structure) design doc provides some useful background information on the structure of resources in a devfile registry (and its Git repository).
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -30,9 +38,50 @@ The following are required to build the devfile index container image containing
...
-    - PR tests will also verify this.
+5) Open a pull request against this repository with a brief description of the change.
+
+
+### Updating
+
+Updating an existing devfile stack is relatively straightforward:
+
+1) Find the stack under the `stacks/` folder that you wish to update
+2) Make the necessary changes to the stack (such as updating image tags, commands, starter projects, etc)
```
jc-berger, line 49: Not a big deal, but parentheses do imply a lack of importance. Consider rephrasing to:

`Make the necessary changes to the stacks, such as: updating image tags, commands, starter projects, etc.`

CONTRIBUTING.md
```diff
@@ -30,9 +38,50 @@ The following are required to build the devfile index container image containing
...
+3) Update the version of stack, following the [semantic versioning format](https://semver.org/)
+4) Test your changes:
+    
+    - Minimally, testing with odo (`odo create`, `odo push`, etc) is recommended, however if your Devfile is used with other tools, it's recommended to test there as well
```
jc-berger, line 53: I read `minimally` a couple times. I think it'd be simpler to say,
`Testing with odo, like `odo create` and `odo push`, is recommended. However, if you use your devfile with other tools, you can test it with those tools.`

CONTRIBUTING.md
```diff
@@ -30,9 +38,50 @@ The following are required to build the devfile index container image containing
...
+    - The
```
jc-berger, line 54: Eliminate this bullet point.

CONTRIBUTING.md
```diff
@@ -30,9 +38,50 @@ The following are required to build the devfile index container image containing
...
-    - For now, strictly an optional requirement, but recommended.
-    - PR tests will also verify this.
+5) Open a pull request against this repository with a brief description of the change.
+
+
+### Updating
+
+Updating an existing devfile stack is relatively straightforward:
+
+1) Find the stack under the `stacks/` folder that you wish to update
```
jc-berger, line 48: Also not a big deal, but we might want to include periods at the end of these bullets/numbered lists. They're all complete sentences, hence, they probably should have periods. _But_, I always considered lists and such as being more flexible with grammar :)

CONTRIBUTING.md
```diff
@@ -1,27 +1,35 @@
...
 
-This document outlines the requirements for contributing a devfile stack to this repository
+This document outlines the requirements for contributing a devfile stack or sample to this repository
 
 The [devfile registry structure](https://github.com/devfile/api/blob/main/docs/proposals/registry/registry-structure.md#repository-structure) design doc provides some useful background information on the structure of resources in a devfile registry (and its Git repository).
 
 ## Prerequisites
 
-The following are required to build the devfile index container image containing your stack:
+The following are required to build the devfile index container image containing your stack or sample:
```
jc-berger, line 9: I think passive voice here reads fine. If we wanted to be extra user focused, we could make it active voice:

`To build the devfile index container image containing your stack or sample, you must complete the following:`

CONTRIBUTING.md
johnmcollier: You `should` test it, rather than `can` test it?

CONTRIBUTING.md
```diff
@@ -1,27 +1,35 @@
...
+
+    - Name: The name of the devfile stack, e.g. `java-springboot`
+    - Display Name: The longer name of your devfile stack, e.g. `Spring Boot®`
+    - Description: A brief description of your devfile stack, e.g. `Spring Boot® using Java`
+    - Version: The version of your stack, in semnatic version format, e.g. `1.0.0`
+
 2) Add a folder for the stack to `stacks/` in this repository.
   
-    - Name format should be in the form of `<language>-<framework>`. E.g. `java-quarkus`, `python-django`, etc.
+    - The name should match the devfile stack's name, and be in the format `<language>-<framework>`. E.g. `java-quarkus`, `python-django`, etc.
```
jc-berger, line 32: _Should_ the name match the devfile stack? Or _must_ it? If the latter, simply say, `The name must match the devfile stack's name` or make it user focused like, `Make sure the name matches the devfile stack's name and be in the format...`

CONTRIBUTING.md
johnmcollier: Huh, TIL

johnmcollier: Fixed

jc-berger (approved):
LGTM from docs point of view. However, I did leave suggestions for improvements, none of which are mandatory changes :)

CONTRIBUTING.md
johnmcollier: You must `install the following` rather than `complete the following`? Since it's highlighting the tools we need to install?

CONTRIBUTING.md
```diff
@@ -30,9 +38,49 @@ The following are required to build the devfile index container image containing
...
+4) Test your changes:
+    
+    - Minimally, testing with odo (`odo create`, `odo push`, etc) is recommended, however if your Devfile is used with other tools, it's recommended to test there as well.
+5) Open a pull request against this repository with your changes.
+
+## Samples
+
+### Contributing
+
+The devfile samples used in this devfile registry are stored in the `extraDevfileEntries.yaml` file in the root of the repository. To add a devfile sample:
```
GeekArthur, line 60: I think the purpose of `extraDevfileEntries.yaml` is for letting user add remote stack/sample to the registry instead of only for adding samples, that is also why we have top level attribute stack/sample in the file. Given that, if we still divide the docs sections into stack and sample, I think we may need to mention the `extraDevfileEntries.yaml` in stack section to add remote stack into the registry.

CONTRIBUTING.md
```diff
@@ -30,9 +38,49 @@ The following are required to build the devfile index container image containing
...
+
+1) Open `extraDevfileEntries.yaml` in an editor
+2) Add an entry to the file similar to the following:
```
GeekArthur, line 63: Should we also let user verify the mandatory field (eg. git) for sample here? Also for the git field, we may also mention that we only support one remote.

GeekArthur (changes requested):

CONTRIBUTING.md
jc-berger: If it's recommended for users to test with the other tools, I then yes, say `should` :)

CONTRIBUTING.md
jc-berger: Sure, `must install the following` sounds great!

CONTRIBUTING.md
johnmcollier: @GeekArthur We don't really support remote stacks at this moment (in terms of registry bootstrapping), so I deliberately left it out.

CONTRIBUTING.md
johnmcollier: I mentioned that all of the fields are mandatory since we don't want to be onboarding a sample that's missing anything. I also updated the text to say that only one git remote is supported.

johnmcollier:
@GeekArthur Coming back to this after a month, mind re-reviewing? Thanks!

GeekArthur (approved):

PR #47, opened 2021-07-09
Add architecture workflow check

maysunfaisal:
Signed-off-by: Maysun J Faisal <maysunaneek@gmail.com>

This PR addresses part of comment https://github.com/devfile/api/issues/244#issuecomment-859732271 and introduces a workflow to check for architecture in stack/sample.

Since devfile architecture is not mandatory, the workflow will not block PRs or show up as an error, but exits with EC 1. If it exits with EC1, github-action bot will comment on the PR with the error msg only for stack/sample files with changes in the PR ie; if a stack devfile changes but it has no arch present, the bot will comment on the PR.


johnmcollier:
@maysunfaisal Haven't looked at this yet, but would be cool to use something like https://github.com/marketplace/actions/create-or-update-comment to have a comment made on the PR if the Arch check fails

maysunfaisal:
@johnmcollier that is nice, let me check it out

.github/workflows/validate-devfiles-minikube.yaml
```diff
@@ -37,3 +37,15 @@ jobs:
...
       run: tests/test.sh odo
+  check_architectures:
+    name: Check Stack and Sample Architectures
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    - name: Download yq
+      run: wget https://github.com/mikefarah/yq/releases/download/v4.9.8/yq_linux_amd64 -O ./yq && chmod +x ./yq
```
johnmcollier, line 48: Nit: We can also get yq using https://github.com/redhat-actions/openshift-tools-installer

tests/check_architectures.sh
```diff
@@ -0,0 +1,57 @@
...
+    fi
+}
+
+YQ_PATH=$1
+if [ -z $YQ_PATH ]; then
+  YQ_PATH=yq
+fi
+
+checkStacks
+checkSamples
```
johnmcollier, line 52: What if the `extraDevfileEntries.yaml` file doesn't exist?

In case this script needs to be used in other repos (like https://github.com/redhat-developer/devfile-registry), should we only call `checkSamples` if the file exists?

.github/workflows/validate-devfiles-minikube.yaml
```diff
@@ -37,3 +37,15 @@ jobs:
...
+    - name: Check stacks and samples
+      continue-on-error: true
```
johnmcollier, line 50: With this field set, is there a scenario in which `check_architectures.sh` could exit with an non-zero code, and we would want this check to still fail?

.github/workflows/validate-devfiles-minikube.yaml
```diff
@@ -37,3 +37,15 @@ jobs:
         odo: "latest"
     - name: Validate the devfile stacks
       run: tests/test.sh odo
+  check_architectures:
+    name: Check Stack and Sample Architectures
+    continue-on-error: true
```
johnmcollier, line 42: Do we need this set here and down below?

johnmcollier (changes requested):
Left some comments.

I assume we're gonna leave the stack architectures empty for now and let the stack owners and them when they decide?

.github/workflows/validate-devfiles-minikube.yaml
maysunfaisal: so curently the script exits with non zero, if we dont put `continue-on-error: true`, all the builds will finish as ❌  which really doesn't look great..

Screen Shot 2021-07-12 at 1 27 33 PM

.github/workflows/validate-devfiles-minikube.yaml
maysunfaisal: updated

tests/check_architectures.sh
maysunfaisal: updated

.github/workflows/validate-devfiles-minikube.yaml
maysunfaisal: updated

maysunfaisal:
Commenting seems to work on my fork https://github.com/maysunfaisal/registry-1/pull/11 but not on public repo with a PR from fork

From the action's README:
Screen Shot 2021-07-13 at 3 01 05 PM


I will try it out

maysunfaisal:
### UPDATE

Initially I used the action https://github.com/marketplace/actions/create-or-update-comment but ran into issues as documented in https://github.com/devfile/registry/pull/47#issuecomment-879326990

I  later found a github blog to workaround this https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ and was able to achieve the result

POC: https://github.com/maysunfaisal/test-actions/pull/17

johnmcollier (approved):
Looks good, nice changes. The comment will be really nice to have on PRs 👍 .

PR #58, opened 2021-08-03
Add icons for wildfly stacks

johnmcollier:
Adds icons for the two wildfly stacks (java-wildfly and java-wildfly-bootable-jar):

Screen Shot 2021-08-03 at 1 45 06 PM

Sourced from: https://design.jboss.org/wildfly/logo/final/wildfly_logomark.svg


johnmcollier:
@ehsavoie Hi Emmanuel, can you let me know if this is the right icon to use for the wildfly stacks, or if there's a more appropriate logo? Thanks!

elsony:
@ehsavoie Would you mind confirming if it is the correct icon?

ehsavoie:
yes it is

PR #59, opened 2021-08-03
Add icon for nodejs stack

johnmcollier:
Adds an icon for the nodejs stack:
Screen Shot 2021-08-03 at 1 47 29 PM

Sourced from https://nodejs.org/static/images/logos/nodejs-new-pantone-black.svg

### DIFF
--- a/stacks/nodejs/devfile.yaml
+++ b/stacks/nodejs/devfile.yaml
@@ -4,6 +4,7 @@ metadata:
   version: 1.0.1
   displayName: Node.js Runtime
   description: Stack with Node.js 14
+  icon: https://nodejs.org/static/images/logos/nodejs-new-pantone-black.svg
   tags: ["NodeJS", "Express", "ubi8"]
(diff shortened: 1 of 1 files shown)


johnmcollier:
@BethGriggs Hi Beth, can you let me know if this is the right icon to use for the nodejs devfile stack or if there's another one that should be used? Thanks!

BethGriggs:
This is the Node.js logo, but it would be worth checking that it will adhere to the trademark policies:
* https://nodejs.org/static/documents/trademark-policy.pdf
* https://openjsf.org/wp-content/uploads/sites/84/2021/01/OpenJS-Foundation-Trademark-Policy-2021-01-12.docx.pdf

deboer-tim:
> This is the Node.js logo, but it would be worth checking that it will adhere to the trademark policies:

Thanks, I have read and confirm we'll follow these.

elsony (approved):
/lgtm

PR #62, opened 2021-08-03
Add environment variables for telemetry

GeekArthur:
This PR adds the environment variables to devfile community registry

Signed-off-by: Jingfu Wang <jingfwan@redhat.com>

### DIFF
--- a/.ci/deploy/devfile-registry.yaml
+++ b/.ci/deploy/devfile-registry.yaml
@@ -69,6 +69,11 @@ objects:
             limits:
               cpu: 100m
               memory: ${DEVFILE_INDEX_MEMORY_LIMIT}
+          env:
+          - name: ENABLE_TELEMETRY
+            value: true
+          - name: REGISTRY
+            value: devfile-community-registry
         - image: ${OCI_REGISTRY_IMAGE}:${OCI_REGISTRY_IMAGE_TAG}
           imagePullPolicy: "${OCI_REGISTRY_PULL_POLICY}"
           name: oci-registry
--- a/.ci/deploy/devfile-registry.yaml
+++ b/.ci/deploy/devfile-registry.yaml
@@ -71,9 +71,9 @@ objects:
               memory: ${DEVFILE_INDEX_MEMORY_LIMIT}
           env:
           - name: ENABLE_TELEMETRY
-            value: true
-          - name: REGISTRY
-            value: devfile-community-registry
+            value: ${ENABLE_TELEMETRY}
+          - name: REGISTRY_NAME
+            value: ${REGISTRY_NAME}
         - image: ${OCI_REGISTRY_IMAGE}:${OCI_REGISTRY_IMAGE_TAG}
           imagePullPolicy: "${OCI_REGISTRY_PULL_POLICY}"
           name: oci-registry
@@ -190,3 +190,11 @@ parameters:
   value: "1"
   displayName: Devfile registry replicas
   description: The number of replicas for the hosted devfile registry service
+- name: ENABLE_TELEMETRY
+  value: false
+  displayName: Devfile registry telemetry enablement
+  description: Option to opt in or opt out devfile registry telemetry
+- name: REGISTRY_NAME
+  value: devfile-community-registry
+  displayName: Devfile registry name
+  description: The registry name that is used as identifier for devfile telemetry
--- a/.ci/deploy/devfile-registry.yaml
+++ b/.ci/deploy/devfile-registry.yaml
@@ -191,7 +191,7 @@ parameters:
   displayName: Devfile registry replicas
   description: The number of replicas for the hosted devfile registry service
 - name: ENABLE_TELEMETRY
-  value: false
+  value: "false"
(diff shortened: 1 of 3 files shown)


.ci/deploy/devfile-registry.yaml
```diff
@@ -69,6 +69,11 @@ objects:
             limits:
               cpu: 100m
               memory: ${DEVFILE_INDEX_MEMORY_LIMIT}
+          env:
+          - name: ENABLE_TELEMETRY
+            value: true
```
johnmcollier, line 74: This field should be be configurable with a parameter under https://github.com/devfile/registry/blob/b05971c7cf9fc52f280dc23690d25bc01511b479/.ci/deploy/devfile-registry.yaml#L156

And the default value should be false.

.ci/deploy/devfile-registry.yaml
```diff
@@ -69,6 +69,11 @@ objects:
...
+          - name: REGISTRY
```
johnmcollier, line 75: What's this environment variable for exactly? I thought we were only going to use one project in segment?

And we should use a more descriptive environment variable name.

johnmcollier (changes requested):

.ci/deploy/devfile-registry.yaml
johnmcollier: It should also be configurable with a parameter

GeekArthur: Yes, only one project for segment and woopra, this is the registry name that is used as identifier for telemetry to know  which registry the data come from. I will update the description.

.ci/deploy/devfile-registry.yaml
johnmcollier: The parameter will be set in AppSRE's Gitlab instance, where we have the deployment configurations for staging and prod

GeekArthur: Got it, thanks

johnmcollier (changes requested):
Thanks for making the changes Jingfu. Two more changes needed:
1) Applying the template gives the following error:
```
johncollier@jcollier-mac hosted-registry % oc process -f devfile-registry.yaml | oc apply -f - 
error: failed to read input object (not a Template?): unable to decode "devfile-registry.yaml": v1.Template.Parameters: []v1.Parameter: v1.Parameter.Value: ReadString: expects " or n, but found f, error found in #10 byte of ...|,"value":false},{"de|..., bigger context ...|ry enablement","name":"ENABLE_TELEMETRY","value":false},{"description":"The registry name that is us|...
```
The `env` field in Kubernetes expects strings for values, so I think `ENABLE_TELEMETRY` being a bool is throwing it off. Surrounding the `false` in double quotes fixed it for me.

2) Can you update the `.ci/openshift_integration.sh` script in this repo to point to this template? It looks like this repo's tests are still using the old template in registry-support (that I should've removed)

GeekArthur:
> Thanks for making the changes Jingfu. Two more changes needed:
> 
>     1. Applying the template gives the following error:
> 
> 
> ```
> johncollier@jcollier-mac hosted-registry % oc process -f devfile-registry.yaml | oc apply -f - 
> error: failed to read input object (not a Template?): unable to decode "devfile-registry.yaml": v1.Template.Parameters: []v1.Parameter: v1.Parameter.Value: ReadString: expects " or n, but found f, error found in #10 byte of ...|,"value":false},{"de|..., bigger context ...|ry enablement","name":"ENABLE_TELEMETRY","value":false},{"description":"The registry name that is us|...
> ```
> 
> The `env` field in Kubernetes expects strings for values, so I think `ENABLE_TELEMETRY` being a bool is throwing it off. Surrounding the `false` in double quotes fixed it for me.
> 
>     1. Can you update the `.ci/openshift_integration.sh` script in this repo to point to this template? It looks like this repo's tests are still using the old template in registry-support (that I should've removed)

For the second change, I assume we trigger the integration test script from this repo so I use relative path `deploy/devfile-registry.yaml` to point to the integration test script, also I notice we don't have route.yaml in this repo so it still use the route.yaml from registry-support repo, another alternative is that copying over route.yaml to this repo.

johnmcollier:
Yeah, use deploy/devfile-registry.yaml to refer to it.

And yeah, feel free to copy over the route.yaml as well.

GeekArthur:
@johnmcollier Comments addressed, please review again.

johnmcollier:
@GeekArthur Sorry Jingfu, looks like I was wrong about the paths. I forgot that the OpenShift CI executes in the root of the repository, so it needs to be `.ci/deploy` for the registry and the route yamls.

GeekArthur:
> @GeekArthur Sorry Jingfu, looks like I was wrong about the paths. I forgot that the OpenShift CI executes in the root of the repository, so it needs to be `.ci/deploy` for the registry and the route yamls.
> 
> Otherwise it looks good!

@johnmcollier Updated, please review again

johnmcollier (approved):

PR #68, opened 2021-08-31
Add workflow for triggering deploys to staging

johnmcollier:
Fixes https://github.com/devfile/api/issues/527

Adds a workflow to receive repository dispatches that will trigger deploys to staging

### DIFF
--- a/.github/workflows/trigger-redeploy.yaml
+++ b/.github/workflows/trigger-redeploy.yaml
@@ -22,9 +22,7 @@ jobs:
     
     - name: Trigger Deploy to Staging
       run: |
-        echo ${{ github.event.client_payload.repo }}
-        echo ${{ github.event.client_payload.sha }}
-        git config --global user.email "jcollier@redhat.com"
-        git config --global user.name "John Collier"
+        git config --global user.email "${{secrets.CI_ROBOT_EMAIL}}"
+        git config --global user.name "${{secrets.CI_ROBOT_NAME}}"
         git commit --allow-empty -m "Trigger deploy based on devfile/${{ github.event.client_payload.repo }}@${{ github.event.client_payload.sha }}"
         git push "https://${{secrets.CI_ROBOT}}:${{secrets.REPO_ACCESS_TOKEN}}@github.com/devfile/registry"
--- a/.github/workflows/trigger-redeploy.yaml
+++ b/.github/workflows/trigger-redeploy.yaml
@@ -10,7 +10,7 @@ name: CI
 
 on:
   repository_dispatch:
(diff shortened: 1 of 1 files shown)


johnmcollier:
@GeekArthur Please review when you get the chance, thanks!

johnmcollier:
An example workflow can be found here: https://github.com/johnmcollier/registry-test-workflow/actions/runs/1194557677

With example commits seen here: https://github.com/johnmcollier/registry-test-workflow/commits/main

.github/workflows/trigger-redeploy.yaml
```diff
@@ -0,0 +1,28 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+name: CI
+
+on:
+  repository_dispatch:
+    types: [trigger-redeploy]
```
maysunfaisal, line 13: I cant find whats triggering this event, should this be build? https://github.com/devfile/registry-support/pull/80/files#diff-003cb479ec4293d6e87970d83b072e03fbaf10e40f47378bc302a54a1ad18ed6R70

maysunfaisal: So registry-support would trigger two workflows? This and the one in https://github.com/devfile/registry/blob/main/.github/workflows/pushimge-next.yaml?

johnmcollier: Oops, that should be a `build`

johnmcollier: Fixed.

maysunfaisal: So would you need this workflow to wait for the image push workflow https://github.com/devfile/registry/blob/main/.github/workflows/pushimge-next.yaml to push the latest image, before triggering a deploy to staging? 

What if the trigger to redeploy staging uses the old image, since registry-support is going to dispatch the event and both these workflows will run in parallel

johnmcollier: Good question: Nope. AppSRE builds their own container image that’s used in the deployment. (quay.io/app-sre/devfile-index)

The registry deployment only runs after AppSRE’s Jenkins pipeline has successfully built and pushed the image.

We actually don’t need the image that’s built in the workflow you linked (and I’ve actually debated removing it)

maysunfaisal (approved):

johnmcollier:
CI failures are unrelated to these changes (which do not touch any stacks or tests), so going to go ahead and force merge.

PR #74, opened 2021-10-08
Update tests for new odo version

johnmcollier:
New version of odo requires that we change the following in the tests:
- Don't pass in devfile name when testing devfiles locally
- Must pass in the starter project name now


josephca (approved):
LGTM

PR #77, opened 2021-10-19
Add Devfiles for .NET 3.1 and 5.0

kadel:
### What does this PR do?:
Add stacks for .NET 3.1 and .NET 5.0

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
all tests expect `java-quarkus` pass.
reported as https://github.com/devfile/api/issues/650 

- [x] Documentation
No

### DIFF
--- a/stacks/dotnet50/devfile.yaml
+++ b/stacks/dotnet50/devfile.yaml
@@ -28,8 +28,10 @@ components:
         value: Debug
       - name: STARTUP_PROJECT
         value: app.csproj
-      - name: "ASPNETCORE_ENVIRONMENT"
+      - name: ASPNETCORE_ENVIRONMENT
         value: Development
+      - name: ASPNETCORE_URLS
+        value: http://*:8080
     endpoints:
     - name: http-8080
       targetPort: 8080
--- a/stacks/dotnetcore31/devfile.yaml
+++ b/stacks/dotnetcore31/devfile.yaml
@@ -29,6 +29,8 @@ components:
         value: app.csproj
       - name: ASPNETCORE_ENVIRONMENT
         value: Development
+      - name: ASPNETCORE_URLS
+        value: http://*:8080
     endpoints:
     - name: http-8080
       targetPort: 8080


stacks/dotnet31/devfile.yaml
```diff
@@ -0,0 +1,50 @@
+schemaVersion: 2.1.0
+metadata:
+  name: dotnet50
+  displayName: .NET 3.1
```
tmds, line 4: `.NET Core 3.1`

stacks/dotnet31/devfile.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+starterProjects:
+  - name: dotnetcore
+    git:
+      checkoutFrom:
+        remote: origin
+        revision: dotnetcore-3.1
+      remotes:
+        origin: https://github.com/redhat-developer/s2i-dotnetcore-ex
+    subDir: app
```
tmds, line 18: What does subDir do, does it trim the repo to this folder?

Often .NET repos have sibling folders next to the app that contain additional code that gets used by the app.

Is this pattern supported by `subDir`?

stacks/dotnet31/devfile.yaml
```diff
@@ -0,0 +1,50 @@
...
+    commandLine: dotnet build -c $CONFIGURATION
+    component: dotnet
+    group:
+      isDefault: true
+      kind: build
+
+- id: run
+  exec:
+    workingDir: ${PROJECTS_ROOT}
+    commandLine: dotnet ./bin/$CONFIGURATION/netcoreapp3.1/app.dll
```
tmds, line 45: `app.dll` is specific to the example application.
This should be a parameter too.
The default value can be derived from the folder containing the application (for the example it is `app`) and adding `.dll` to it.

stacks/dotnet50/devfile.yaml
```diff
@@ -0,0 +1,51 @@
+schemaVersion: 2.1.0
+metadata:
+  name: dotnet50
+  displayName: .NET 5.0
```
tmds, line 4: note: `Core` doesn't need to be added to `.NET 5.0`. This is good as-is.

stacks/dotnet31/devfile.yaml
```diff
@@ -0,0 +1,50 @@
+schemaVersion: 2.1.0
+metadata:
+  name: dotnet50
```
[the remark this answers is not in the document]
kadel: oh, copy/paste error :innocent: thx for catching this

stacks/dotnet31/devfile.yaml
tmds: We should remove `subdir` here, and replace it with a parameter on the component.

The parameter can be called `STARTUP_PROJECT`/`PROJECT`.

In our s2i builder it is called `DOTNET_STARTUP_PROJECT` (see https://github.com/redhat-developer/s2i-dotnetcore/blob/master/5.0/build/README.md#environment-variables).

kadel: But the application source code is in the `app` forlder no? In the root of the repo there is just documentation.

If I add something like `DOTNET_STARTUP_PROJECT` I will have to set its default value to `app` to make the starter project work out of the box. And this will mean that every time someone will want to use this devfile with a project where the application in root of the repository they will have to modify `DOTNET_STARTUP_PROJECT`

tmds: > But the application source code is in the app forlder no? In the root of the repo there is just documentation.

Yes, for `s2i-dotnetcore-ex` it is. This is basic `hello` world web app. It doesn't have library projects.

> If I add something like DOTNET_STARTUP_PROJECT I will have to set its default value to app to make the starter project work out of the box. And this will mean that every time someone will want to use this devfile with a project where the application in root of the repository they will have to modify DOTNET_STARTUP_PROJECT

.NET apps rarely live in the repository root, so something needs to be set anyhow. Instead of `subdir`, the user sets `STARTUP_PROJECT`.

stacks/dotnet31/devfile.yaml
tmds: Yes, in our s2i image we also derive it from the project file name.
The project file can be for C# (`.csproj`), it may also be for F# (`.fsproj`) or VB (`vbproj`).

`STARTUP_PROJECT` can be set to a `.??proj` file, but the s2i builder also allows the user to specify its parent folder, and then looks for the project file in the folder. I don't know how smart you can make this `Devfile`.

stacks/dotnet31/devfile.yaml
tmds: I'm going to make some time still this week to get some first hand experience using `odo` with `Devfiles`.

kadel: @tmds I've added a bit of "bash magic" into the commands, so if the `STARTUP_PROJECT` is not set, it will automatically pick first *.csfile or *.fsfile. Not sure if it is good idea or not. If you think that it might be confusing, I can remove it, and we can rely only on `STARTUP_PROJECT`

tmds: > I've added a bit of "bash magic" into the commands, so if the STARTUP_PROJECT is not set, it will automatically pick first *.csfile or *.fsfile. Not sure if it is good idea or not. If you think that it might be confusing, I can remove it, and we can rely only on STARTUP_PROJECT

Let's remove it. The user can easily figure out how to use `STARTUP_PROJECT`. Not all users will grok the bash magic.

tmds:
@kadel :+1: this is looking good. We're getting near the finish.

PR #93, opened 2022-02-15
add missing DEBUG_PORT env variable

kadel:
### What does this PR do?:
Debug commands are using `DEBUG_PORT` variable but that variable is not defined in containers.
This was not an issue for `odo` as it automatically injects this variable if it is not defined, but this creates problems for other tools.

fixes one of the items in https://github.com/devfile/api/issues/679

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [x] Documentation

### DIFF
--- a/stacks/java-maven/devfile.yaml
+++ b/stacks/java-maven/devfile.yaml
@@ -25,6 +25,9 @@ components:
       volumeMounts:
         - name: m2
           path: /home/user/.m2
+      env:
+        - name: DEBUG_PORT
+          value: "5858"
   - name: m2
     volume: {}
 commands:
--- a/stacks/java-openliberty-gradle/devfile.yaml
+++ b/stacks/java-openliberty-gradle/devfile.yaml
@@ -48,6 +48,9 @@ components:
           name: ep1
           targetPort: 9080
           protocol: http
+      env:
+        - name: DEBUG_PORT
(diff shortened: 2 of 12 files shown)


kadel:
java-quarkus is failing due to https://github.com/devfile/api/issues/773

elsony (commented):
Does setting the debug port to the same port means we are increasing the chances of getting port conflict when running different types of applications at the same time? It is fairly common in tools, e.g. IDE, to have configurable debug port numbers, usually auto-generated. Do we need a debug port to explicitly set in every devfile?

kadel:
> Does setting the debug port to the same port means we are increasing the chances of getting port conflict when running different types of applications at the same time? It is fairly common in tools, e.g. IDE, to have configurable debug port numbers, usually auto-generated. Do we need a debug port to explicitly set in every devfile?

If we don't set it in the devfile then the tools will have to set it themself. 
Currently, there is nothing in devfile spec that defines that tools should set this variable.
Odo sets it if it is not defined in the devfile, but Che doesn't so without setting it explicitly in devfile debug in Che won't work (https://github.com/devfile/api/issues/777)

I don't think that this can create conflicts. Each devfile runs as a separate Deployment/Port/Container.

elsony (approved):
/lgtm

PR #106, opened 2022-04-19
Upgrading to WildFly 26.1

ehsavoie:
Upgrading the devFile to latest WildFly release


johnmcollier:
@ehsavoie Could you ammend your commit to include a `Signed-off-by: ...` header? So that the DCO check passes?

Thanks!

PR #118, opened 2022-08-02
feat: Adding stack with minimal Devfile that only contains metadata fields and the schema version

ibuziuk:
Signed-off-by: Ilya Buziuk <ibuziuk@redhat.com>

### What does this PR do?:
Adding stack with minimal Devfile that only contains metadata fields and the schema version.
For Eclipse Che / Dev Spaces that devfile can be used for starting a workspace without any projects cloned. 

Related to https://github.com/eclipse/che/issues/21554

### PR acceptance criteria:

- [ ] Contributing guide
- [ ] Test automation
- [ ] Documentation
- [ ] Check Tools Provider

The build is currently failing with `Error: failed to generate index struct: /build/stacks/empty devfile is not valid: [metadata.language is not set metadata.projectType is not set]` but the question is do we really want to enforce the language and the projectType ?

### DIFF
--- a/stacks/empty/devfile.yaml
+++ b/stacks/empty/devfile.yaml
@@ -1,0 +1,7 @@
+schemaVersion: 2.2.0
+metadata:
+  name: empty
+  version: 1.0.0
+  displayName: Empty Devfile
+  description: Stack with minimal Devfile that only contains metadata fields and the schema version
+  icon: https://raw.githubusercontent.com/devfile/devfile.github.io/master/images/devfileLogo.svg
--- a/stacks/empty/OWNERS
+++ b/stacks/empty/OWNERS
@@ -1,0 +1,5 @@
+# See the OWNERS docs: https://go.k8s.io/owners
+
+reviewers:
+- kadel
+- l0rd


kadel:
Would this stack be ever useful for someone?
This looks a "hack" just for Che :-/

ibuziuk:
> Would this stack be ever useful for someone?
This looks like a "hack" just for Che :-/

This is not meant to be a hack by any means, but rather the most minimal devfile anyone could start with. The analogy could be docker scratch - https://hub.docker.com/_/scratch
In general, we are constantly getting feedback that defiles are complex and hard to start with. The idea is to provide the minimalistic empty / scratch devfile that anyone could start with without overhead @slemeur @l0rd wdyt?

stacks/empty/devfile.yaml
```diff
@@ -0,0 +1,7 @@
+schemaVersion: 2.2.0
+metadata:
+  name: empty
```
ibuziuk, line 3: might be better to name it as scratch to use the common analogy in the container world with https://hub.docker.com/_/scratch

kadel:
>This is not meant to be a hack by any means, but rather the most minimal devfile anyone could start with. The analogy could be docker scratch - https://hub.docker.com/_/scratch

This is a good analogy. Similar to a scratch container, this just on its own is not useful for anything. But it is a good starting point for building on top of it. That is why I don't think that this should be in a devfile registry together with other devfile stacks.
Something like this should be mentioned in the documentation together with description how to bould your own devfile stack. 


>In general, we are constantly getting feedback that defiles are complex and hard to start with. The idea is to provide the minimalistic empty / scratch devfile that anyone could start with without overhead @slemeur @l0rd wdyt?


Devfile is, by its own definition, complex and hard to start with, and I'm afraid that in its current form, it always will be. Devfile is just a big complex structure.

The approach to address this should be that we try to create a good set of devfile stacks that can be used out-of-the box for most of the projects. The only time user would need to touch devfile.yaml directly is if the users' project has something non-standard, unique, or specific to only their project.

michael-valdron:
This PR has been open for almost 2 years now and will close for now, feel free to re-open if you'd like to resume effort on this contribution.

cc @devfile/devfile-services-team @ibuziuk

PR #124, opened 2022-09-21
Use proper API version for templates

johnmcollier:
Need to use template.openshift.io/v1 instead of v1 for OpenShift templates


johnmcollier:
Force merging is required here

PR #134, opened 2022-10-04
Updated the devfile metadata to be more consistent

schultzp2020:
### What does this PR do?:

1. Updates the devfile metadata to be more consistent i.e. capitalize Java in the language property.
2. Adds editorconfig and gitattributes
3. Adds github action concurrency check

N/A

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [ ] Documentation
- [x] Check Tools Provider


schultzp2020:
Please review your stacks. Thanks!
@kadel 
@ajm01 
@maxandersen 
@ehsavoie 
@BethGriggs

ehsavoie (approved):
For java-wildfly-bootable and java-wildfly

schultzp2020:
@maxandersen and @BethGriggs please review my changes before its planned merge date of 2022/10/13.

extraDevfileEntries.yaml
```diff
@@ -15,9 +17,11 @@ samples:
     displayName: Basic Quarkus
     description: A simple Hello World Java application using Quarkus
     icon: https://design.jboss.org/quarkus/logo/final/SVG/quarkus_icon_rgb_default.svg
-    tags: ['Java', 'Quarkus']
-    projectType: quarkus
-    language: java
+    tags:
+      - Java
+      - Quarkus
+    projectType: Quarkus
```
maxandersen, line 23: is the type not still lower case `quarkus`?

stacks/java-openliberty/devfile.yaml
```diff
@@ -14,24 +14,30 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 schemaVersion: 2.1.0
 metadata:
   name: java-openliberty
-  version: 0.8.1
-  displayName: 'Open Liberty Maven'
+  version: 0.8.2
```
scottkurz, line 21: What would be the downside of leaving the version as-is for the four Liberty (Open/WebSphere , Maven/Gradle) devfiles?

We started down a path of using our own versioning strategy integrating our outer loop artifacts.   Following that to its conclusion we'd want to make other changes.

But if this change is just syntax cleanup, well, I think we could say this is equivalent to our previous version, e.g. 0.8.1 in this case, and merge in the change.   

(And we could open an issue in our devfile source repo to make the same changes there:  https://github.com/OpenLiberty/devfile-stack/issues ).   WDYT?

scottkurz (changes requested):
Do we need to bump version?

extraDevfileEntries.yaml
schultzp2020: `Language` and `projectType` should follow the same case as the commercial display name.

stacks/java-openliberty/devfile.yaml
elsony: In general, we recommend version updates (semver) when the stack is updated so that users know the underlying stack has been changed. Having said that, it is just a recommendation so it is up to the stack owner to have the final decision on the version number.

scottkurz: I'd rather just leave the version alone in this PR then.

I opened: https://github.com/OpenLiberty/devfile-stack/issues/212 to follow-up and make these changes.  We can still view that repo as the source of record of our devfile.

But I think it's OK to go ahead and merge the rest of the changes.  We might not get to it for a few weeks so don't want to keep you waiting.   

There's an opportunity for minor user confusion here, I know, but unless there's some specific reason (like some version of some devfile tooling having a bug in the older syntax) then I don't think it will end up mattering all that much.

johnmcollier (approved):

kadel:
The tests with odo v3 were failing, but it still got merged :-(

Now the dotnet devfiles are not working with odo.
This is a bug in odo (https://github.com/redhat-developer/odo/issues/6234), but still, this is precisely the reason we have tests here, so we know when something gets broken, and we can fix it.

If we ignore failing tests then I don't know why we even have them.

johnmcollier:
@kadel Hey, sorry about that, you’re absolutely right, and merging this is in was premature. We had discussed this PR on a team call yesterday,, and were under the impression it would be safe to merge in, but that was clearly a mistake.

It also appears that the odo v3 tests are not a required check to merge, and that should definitely change.

PR #136, opened 2022-10-05
Add "with a Devfile" to descriptions.

tmds:
This makes it more clear what the differences is with other options in the OpenShift developer catalog for the same language/runtime.

@kadel ptal.

### DIFF
--- a/extraDevfileEntries.yaml
+++ b/extraDevfileEntries.yaml
@@ -2,7 +2,7 @@ schemaVersion: 1.0.0
 samples:
   - name: nodejs-basic
     displayName: Basic Node.js
-    description: A simple Hello World Node.js application
+    description: A simple Hello World Node.js application with a Devfile
     icon: https://nodejs.org/static/images/logos/nodejs-new-pantone-black.svg
     tags: ['NodeJS', 'Express']
     projectType: nodejs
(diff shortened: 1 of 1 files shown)


tmds:
> This makes it more clear what the differences is with other options in the OpenShift developer catalog for the same language/runtime.

For example:

dev catalog

elsony:
Adding "with a Devfile" on the entries in the devfile registry looks redundant since those entries are from a devfile registry so it is obvious that the samples contain a devfile. In OpenShift developer catalog, there is already the type on the top right of each card to describe the type, e.g. Devfiles, Builder Images, Helm Charts. Is that clear enough already that it is a devfile sample?

tmds:
> there is already the type on the top right of each card

I think it helps users pick between these options by being more descriptive about the difference.

This is similar to how the `Helm Chart` says: _A Helm chart to ..._.

elsony:
Those entries are doing a live query on the public community registry: https://registry.devfile.io/
It will be strange when viewing the public community registry to have every entry have "with a devfile" at the end with is already implied.

johnmcollier:
Yeah, since we need to display these in places outside of the OpenShift Console, I agree with Elson.

tmds:
Another example, where it's even less clear one option is for using Helm:

image

elsony:
Maybe the UI is missing the type information in that case, e.g. in the All services catalog, we have:
image
which will be clear enough without mentioning helm or devfile in the description.

tmds:
The screenshot I shared is what you get when clicking 'View all samples':

image

> Maybe the UI is missing the type information in that case

Yes, the information is missing in this view.

elsony:
I have created https://issues.redhat.com/browse/ODC-7198 to request adding the technology type information to the samples view.

PR #153, opened 2022-12-12
Updated composer image

schultzp2020:
Signed-off-by: Paul Schultz <pschultz@pobox.com>

### What does this PR do?:

Updates composer to the latest image

https://github.com/devfile/api/issues/997

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [ ] Documentation
- [x] Check Tools Provider

### DIFF
--- a/stacks/php-laravel/devfile.yaml
+++ b/stacks/php-laravel/devfile.yaml
@@ -1,9 +1,10 @@
 schemaVersion: 2.1.0
 metadata:
-  description: 'Laravel is an open-source PHP framework, which is robust and easy to understand.
+  description:
+    "Laravel is an open-source PHP framework, which is robust and easy to understand.
     It follows a model-view-controller design pattern.
     Laravel reuses the existing components of different frameworks which helps in creating a web application.
-    The web application thus designed is more structured and pragmatic.'
+    The web application thus designed is more structured and pragmatic."
   displayName: Laravel
   icon: https://raw.githubusercontent.com/devfile-samples/devfile-stack-icons/main/laravel.svg
   name: php-laravel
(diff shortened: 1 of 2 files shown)


stacks/php-laravel/devfile.yaml
```diff
@@ -27,7 +28,7 @@ components:
       endpoints:
         - name: http-laravel
           targetPort: 8000
-      image: quay.io/devfile/composer:2.1.11
+      image: quay.io/devfile/composer:latest
```
johnmcollier, line 31: Is it possible to use something other than latest. Something like `composer:2` or `composer:2.4` The reason we changed the composer image away from latest was that it was bumped to a PHP version that broke the starter projects in the devfile, and I'd like to avoid that issue happening again the next time PHP is updated.

johnmcollier: Also, stack version should be bumped

johnmcollier (changes requested):

stacks/php-laravel/devfile.yaml
kadel: Also, this image has [several critical vulnerabilities](https://quay.io/repository/devfile/composer/manifest/sha256:110fff023c715eb30fc8a2f4806b5b966cfff5c6104d1396ab73c78a3524083d?tab=vulnerabilities) :-(

johnmcollier: > We could probably limit it to composer:2; however, we don't do this for any other stacks. We also have tests that would detect if it failed.

Yes, but ideally we can avoid getting broken in the first place.

kadel: >Sadly there isn't anything we can do about this because RH doesn't have a composer image.

Who maintains this image?

johnmcollier (approved):

PR #169, opened 2023-03-17
[java-springboot] Use LTS JDK 17 image

rm3l:
### What does this PR do?:
This updates the container image to `registry.access.redhat.com/ubi9/openjdk-17`, to keep up with the latest JDK LTS, security-wise.
Also JDK 17 is now a requirement for recent versions of Spring Boot.

&mdash;

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [ ] Documentation
- [x] Check Tools Provider
Tested with `odo`.

### DIFF
--- a/stacks/java-springboot/1.3.0/devfile.yaml
+++ b/stacks/java-springboot/1.3.0/devfile.yaml
@@ -1,0 +1,65 @@
+schemaVersion: 2.1.0
+metadata:
+  name: java-springboot
+  displayName: Spring Boot®
+  description: Java application using Spring Boot® and OpenJDK 11
+  icon: https://raw.githubusercontent.com/devfile-samples/devfile-stack-icons/main/spring.svg
+  tags:
+    - Java
+    - Spring
+  projectType: springboot
+  language: Java
+  version: 1.3.0
+  globalMemoryLimit: 2674Mi
+starterProjects:
+  - name: springbootproject
+    git:
+      remotes:
+        origin: "https://github.com/odo-devfiles/springboot-ex.git"
+components:
+  - name: tools
+    container:
+      image: registry.access.redhat.com/ubi9/openjdk-17:1.16-1.1696518670
+      command: ["tail", "-f", "/dev/null"]
+      memoryLimit: 768Mi
+      mountSources: true
+      endpoints:
+        - name: http-springboot
+          targetPort: 8080
+        - exposure: none
+          name: debug
+          targetPort: 5858
+      volumeMounts:
+        - name: m2
+          path: /home/user/.m2
+      env:
+        - name: DEBUG_PORT
+          value: "5858"
+  - name: m2
+    volume:
+      size: 3Gi
+commands:
+  - id: build
+    exec:
+      component: tools
+      workingDir: ${PROJECT_SOURCE}
+      commandLine: "mvn clean -Dmaven.repo.local=/home/user/.m2/repository package -Dmaven.test.skip=true"
+      group:
+        kind: build
+        isDefault: true
+  - id: run
+    exec:
+      component: tools
+      workingDir: ${PROJECT_SOURCE}
+      commandLine: "mvn -Dmaven.repo.local=/home/user/.m2/repository spring-boot:run"
+      group:
+        kind: run
+        isDefault: true
+  - id: debug
+    exec:
+      component: tools
+      workingDir: ${PROJECT_SOURCE}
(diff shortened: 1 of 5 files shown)


kadel:
Now that we have multi-version support, it might be better to keep the old versions (1.2.0 and 2.0.0) there

rm3l:
> Now that we have multi-version support, it might be better to keep the old versions (1.2.0 and 2.0.0) there

Done. I hesitated to do so in the beginning mainly because of the maintenance overhead, but yes, it makes sense to keep the old versions. The default version of the stack still points to the old 1.2.0, but let me know if you think this should be changed.

kadel:
> The default version of the stack still points to the old 1.2.0, but let me know if you think this should be changed.

I would change it, but this is also a question or @elsony

kim-tsao:
> > The default version of the stack still points to the old 1.2.0, but let me know if you think this should be changed.
> 
> I would change it, but this is also a question or @elsony

I raised this in a meeting and we agreed we need to update our samples to work with parent v2.2 versions (currently they don't).   This is the tracking issue: https://github.com/devfile/api/issues/1112

mohitsuman:
@elsony @kim-tsao can we please get this PR merged? We have customers whose workflow is broken due to the older JDK version. We would need this on priority.

rm3l:
> > > The default version of the stack still points to the old 1.2.0, but let me know if you think this should be changed.
> > 
> > 
> > I would change it, but this is also a question or @elsony
> 
> I raised this in a meeting and we agreed we need to update our samples to work with parent v2.2 versions (currently they don't so if we change the default they will break). This is the tracking issue: [devfile/api#1112](https://github.com/devfile/api/issues/1112)

@kim-tsao @elsony @kadel https://github.com/devfile/api/issues/1112 is mostly about parent Devfiles using v2.2.0 schema. But this PR is mostly about updating the container image.
I think we can update the default to 1.3.0 here (still using Devfile 2.1.0, but with JDK 17), so we are not blocked on https://github.com/devfile/api/issues/1112. And once https://github.com/devfile/api/issues/1112 is merged, we would be able to switch the default to the version using Devfile 2.2.0 (maybe as part of https://github.com/devfile/api/issues/1129). WDYT?

thepetk:
> so we are not blocked on [devfile/api#1112](https://github.com/devfile/api/issues/1112)

@rm3l 

Fyi after the implementation of https://github.com/devfile/api/issues/1209 we are able to have multiple versions of samples and add them to `extraDevFileEntries.yaml`. For this specific stack we have already an opened PR on the springboot sample here: https://github.com/devfile-samples/devfile-sample-java-springboot-basic/pull/33

rm3l:
> > so we are not blocked on [devfile/api#1112](https://github.com/devfile/api/issues/1112)
> 
> @rm3l
> 
> I'm not sure I have the full context for the usage of multiple sample versions for this case.
> 
I think the initial point was whether to switch the default version of this stack to the latest one here (which would be using Devfile 2.2.0). However, the problem mentioned by Kim is that doing so would break the current samples.

> But, fyi after the implementation of [devfile/api#1209](https://github.com/devfile/api/issues/1209) we are able to have multiple versions of samples and add them to `extraDevFileEntries.yaml`. For this specific stack we have already an opened PR on the springboot sample here: [devfile-samples/devfile-sample-java-springboot-basic#33](https://github.com/devfile-samples/devfile-sample-java-springboot-basic/pull/33)

Ok (correct me if I'm wrong), so https://github.com/devfile/api/issues/1209 paved the way to supporting versioned samples, but the actual work of updating [`extraDevfileEntries.yaml`](https://github.com/devfile/registry/blob/main/extraDevfileEntries.yaml) would be done in https://github.com/devfile/api/issues/1112, right?

To move forward with this PR, I'm suggesting we decorrelate it from the work being done on samples, by updating the default version of this stack to the one using Devfile 2.1.0 but with JDK 17. This would keep backward compatibility with the current samples as I guess they would still be working with parent v2.1.0 (only with a different container image).
And the task of updating `extraDevFileEntries` would be handled separately as part of https://github.com/devfile/api/issues/1112, at which point we could also consider switching the default stack version to a version on Devfile 2.2.0 (scope of https://github.com/devfile/api/issues/1129).

rm3l:
As agreed with the rest of the team, I'm switching the default version of this stack to 1.3.0 (the one still on Devfile 2.1.0, but with a more recent JDK 17), for backward compatibility with the samples.

Also rebased and force-pushed as the branch was a bit outdated.

kadel (approved):

PR #170, opened 2023-03-17
[go] Add debug command to Devfile

rm3l:
### What does this PR do?:
This adds a missing `debug` command to the Go stack.

&mdash;

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [ ] Documentation
- [x] Check Tools Provider
Checked with `odo`.

Running `odo dev --debug` and attaching a remote debugger to the local port forwarded by odo should work:

image

### DIFF
--- a/stacks/go/1.1.0/devfile.yaml
+++ b/stacks/go/1.1.0/devfile.yaml
@@ -1,0 +1,76 @@
+schemaVersion: 2.1.0
+metadata:
+  name: go
+  displayName: Go Runtime
+  description: Go (version 1.18.x) is an open source programming language that makes it easy to build simple, reliable, and efficient software.
+  icon: https://raw.githubusercontent.com/devfile-samples/devfile-stack-icons/main/golang.svg
+  tags:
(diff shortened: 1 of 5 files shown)


kadel:
I would keep the old versions there as they are and added new versions with the changes.

rm3l:
> I would keep the old versions there as they are and added new versions with the changes.

Done. I hesitated to do so in the beginning mainly because of the maintenance overhead, but yes, it makes sense to add new versions. I haven't changed the default version for now.

kim-tsao:
> > I would keep the old versions there as they are and added new versions with the changes.
> 
> Done. I hesitated to do so in the beginning mainly because of the maintenance overhead, but yes, it makes sense to add new versions. I haven't changed the default version for now.

Yes, let's keep the default versions as-is for now.  The Devfile samples may be dependent on them as parents and may affect parsing when moving up to a newer parent version (as I've encountered).  We will need to perform additional testing and/or modify the samples to get them to work.

PR #177, opened 2023-04-06
Adding a devfile for the Universal Developer Image

ibuziuk:
### What does this PR do?:
Adding a devfile for the Universal Developer Image

https://github.com/eclipse/che/issues/22126

### PR acceptance criteria:

- [ ] Contributing guide
- [ ] Test automation
- [ ] Documentation
- [ ] Check Tools Provider

- build is failing due to the fact that `metadata.projectType` is not set. Not sure why the field is required though, according to the spec it is optional
- tested against workspaces.openshift.com via https://raw.githubusercontent.com/ibuziuk/registry/udi_devfile/stacks/udi/devfile.yaml (empty workspace is created, 'tools' is provisioned)

image

- need to add an icon for the stack, can we potentially reuse devfile icon?

### DIFF
--- a/stacks/udi/devfile.yaml
+++ b/stacks/udi/devfile.yaml
@@ -1,0 +1,29 @@
+schemaVersion: 2.2.0
+metadata:
+  name: udi
+  displayName: Universal Developer Image
+  description: Universal Developer Image provides various programming languages tools and runtimes for instant coding
+  icon: https://landscape.cncf.io/logos/devfile.svg
+  tags:
+    - Java
+    - Maven
+    - Scala
+    - PHP
+    - .NET
+    - Node.js
+    - Go
+    - Python
+    - Pip
+    - ubi8
+  projectType: universal
+  language: Polyglot
+  version: 1.0.0
+components:
+  - name: tools
+    container:
+      image: quay.io/devfile/universal-developer-image:ubi8-latest
+      memoryLimit: 6G
+      memoryRequest: 512Mi
+      cpuRequest: 1000m
+      cpuLimit: 4000m
+      mountSources: true
--- a/stacks/udi/OWNERS
+++ b/stacks/udi/OWNERS
@@ -1,0 +1,5 @@
+# See the OWNERS docs: https://go.k8s.io/owners
+
+reviewers:
+- ibuziuk
+- l0rd


johnmcollier:
> build is failing due to the fact that metadata.projectType is not set. Not sure why the field is required though, according to the spec it is optional

The field isn't required in the spec, but the registry viewer requires it currently

ibuziuk:
@johnmcollier is there some generic value we can use for UDI? e.g. 'default' ?

elsony:
@thepetk Can you test out to see if the existing devfile matching from Alizer will continue to match the existing language-specific stacks/samples as the suggested devfile entries?

stacks/udi/devfile.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - Python
+    - Pip
+    - ubi8
+  projectType: default
+  language: Polyglot
+  version: 1.0.0
+components:
+  - name: tools
+    container:
+      image: quay.io/devfile/universal-developer-image:ubi8-latest
```
elsony, line 24: We usually recommend a specific version of the image instead of "latest" for stability reasons. Just in case the latest image got changed that affects the function of the devfile. Is there a specific version that we can use?

ibuziuk: @l0rd wdyt, as I recall we want to stick to latest

stacks/udi/devfile.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - Maven
+    - Scala
+    - PHP
+    - .NET
+    - Node.js
+    - Go
+    - Python
+    - Pip
+    - ubi8
+  projectType: default
```
kadel, line 18: what about `universal`, wouldn't it be better?

stacks/udi/devfile.yaml
kadel: This image has a lot of things in different versions and in some cases there can be even multiple versions of the same thing. 
A versioning image beyond UBI8 probably wouldn't make much sense. I think that it is actually good to have `latest` in this case. As I understand UDI users are expected to manage versions themself using things like SDKMan, nvm, virtuelenv, etc....

stacks/udi/devfile.yaml
ibuziuk: I think `default` is better since it is neutral from the semantics perspective

kadel: In every other Devfile this is used to indicate language or framework, `default` looks weird to me.
Maybe my bigger problem with using `default` is that it could indicate that users use this Devfile **by default**, which I have many problems with. 

This devfile with UDI can still be helpful. But it should not be presented as a **default** Devfile.  I think that primarily we need to guide users towards using language or frameworks-specific Devfiles.  This universal devfile should be considered only if no other Devfile stack can be used or if the project uses multiple languages.

kadel:
BTW: Devfile without a single command doesn't make much sense from `odo` perspective. 

This is why `odo` currently fails with devfile like this. 
```
▶ odo dev
 ✗  no command of kind "run" found in the devfile
```

This will be addressed on `odo`'s side in 

- https://github.com/redhat-developer/odo/issues/6731
- https://github.com/redhat-developer/odo/issues/6568


---

This stack will have to be skipped in odo v3 tests by adding `--skip="stack: udi" \`to https://github.com/devfile/registry/blob/26a9a786a3632478563af066cfe2996e1efbfad4/tests/check_odov3.sh#L25

---

`odo` v3 tests on this PR should be failing. 
It looks like the tests are still not running correctly, probably due to https://github.com/devfile/api/issues/1097 (cc @mike-hoang)

kadel:
Another thing to note here is that in order to fully support UDI in odo we will require builds for amd64 (https://github.com/devfile/api/issues/1102)

stacks/udi/devfile.yaml
kadel: > What if there are multiple languages and frameworks used in a particular devfile?
In general, I believe the projectType field should be optional, I do not understand why it is enforced at the moment on the registry end.

I agree with this. It is really weird that the registry requires something that is marked as optional in the Devfile spec. (cc: @elsony, @johnmcollier)


> this is about projectType which has nothing to do with devfile being the default per se, right?
universal is a weird name for projectType since it is very ambiguous

I was thinking about using `universal` as it is also in the image name.

The best would be to completely omit that, but we can do that :-(

We could keep `default` for now, fix the registry to not require it, and then remove it. But I know how this is going to play out; the change will never come and we will get stuck with `default` forever :-(

elsony: +1 on changing it to `universal`. I think `default` is a bit deceiving as well since it implies being the best of all. We require that as part of the registry since we want consistency in the entries in the registry. There are other fields that are optional in the spec but required for the onboarding to the community registry for the same reasons, e.g. icons, description, etc.

stacks/udi/devfile.yaml
elsony: ok, we can leave it as latest.

ibuziuk:
@kadel @elsony folks, is there any other action required before merging this PR?

stacks/udi/devfile.yaml
ibuziuk: changed to `universal` even though I do not like the approach that registry enforces certain metadata that is odo specific

PR #198, opened 2023-08-11
rhtap-tests: Remove `-` from namespace generation

flacatus:


thepetk (approved):
/lgtm after all tests have passed :) The dash is already added here: https://github.com/redhat-appstudio/e2e-tests/blob/main/pkg/utils/util.go#L204

PR #208, opened 2023-09-29
add lifecycle doc

kim-tsao:
### What does this PR do?:

Documents a proposed devfile lifecycle management process

_Link to github issue(s)_

https://github.com/devfile/api/issues/1066

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [x] Documentation
- [x] Check Tools Provider


LIFECYCLE.md
```diff
@@ -0,0 +1,55 @@
+# Lifecycle Maintenance of Devfile Registry Stacks
+
+The purpose of this document is to clarify the roles and responsibilities for maintaining the Devfile stacks and/or samples present in both the [Devfile Community registry](https://github.com/devfile/registry) and [Red Hat product registry](https://github.com/redhat-developer/devfile-registry).
+Devfiles are intended to be used by developers to build their applications with tooling clients that support the devfile spec, so it’s important that stack providers avoid introducing breaking changes or allow security vulnerabilities to go unremediated which would result in a degradation of our end users applications.  In short, we need to ensure our devfiles are trusted enough to be used and in order to achieve this, there is a level of shared responsibility by all of the roles defined below.
```
thepetk, line 4:
```suggestion
Devfiles are intended to be used by developers to build their applications with tooling clients that support the devfile spec. As a result, it’s important that stack providers avoid introducing breaking changes or allowing security vulnerabilities to go unremediated which would result in a degradation of our end users applications.  In short, we need to ensure our devfiles are trusted enough to be used. In order to achieve this, there is a level of shared responsibility in all of the roles defined below.
```

LIFECYCLE.md
```diff
@@ -0,0 +1,55 @@
...
+| `Devfile Team`          | The team that is responsible for onboarding vendor devfiles for both the community and product registries.                              |
+| `Stack/Sample Provider` | The person or team that is responsible for developing the devfile stack and/or that is available for sharing on our devfile registries. |
+| `Tooling Clients`       | These are the tools that support building devfile based applications                                                                    |
+| `End Users`             | Application developers that consume devfiles for the purpose of building their own custom applications                                  |
+
+The following sections describe the steps that can happen in the lifecycle of a devfile.
+
+## [Onboarding](https://github.com/redhat-developer/devfile-registry/blob/main/CONTRIBUTING.md)
+When a `Stack Provider` is ready to share their devfile to the public/product registry, they must follow the steps in the contributing guide to ensure there’s basic information that identifies the version, owner, description, etc of the stack and ensure the stack meets minimal testing requirements against the supported clients.
```
thepetk, line 17:
```suggestion
When a `Stack Provider` is ready to share their devfile to the public/product registry, they must follow the steps in the [contributing guide](CONTRIBUTING.md) to ensure there’s basic information that identifies the version, owner, description, etc of the stack and ensure the stack meets minimal testing requirements against the supported clients.
```

thepetk (changes requested):
I've added some minor suggestions. Feel free to reject them if you think they are not correct :)

A qq I have is about versioning. Should we include a section in the lifecycle doc about versioning or you think is un related (e.g how a stack provider should be versioning their stack's versions etc)

LIFECYCLE.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+| Role                    | Description                                                                                                                             |
+|:------------------------|:----------------------------------------------------------------------------------------------------------------------------------------|
+| `Devfile Team`          | The team that is responsible for onboarding vendor devfiles for both the community and product registries.                              |
+| `Stack/Sample Provider` | The person or team that is responsible for developing the devfile stack and/or that is available for sharing on our devfile registries. |
+| `Tooling Clients`       | These are the tools that support building devfile based applications                                                                    |
```
rm3l, line 11:
```suggestion
| `Tooling Clients`       | These are the tools that support building devfile-based applications                                                                    |
```

LIFECYCLE.md
```diff
@@ -0,0 +1,55 @@
...
+| `End Users`             | Application developers that consume devfiles for the purpose of building their own custom applications                                  |
+
+The following sections describe the steps that can happen in the lifecycle of a devfile.
+
+## [Onboarding](https://github.com/redhat-developer/devfile-registry/blob/main/CONTRIBUTING.md)
```
rm3l, line 16: I guess this should link to the `CONTRIBUTING.md` file in this repo, not the RH Product registry, no?

```suggestion
## [Onboarding](CONTRIBUTING.md)
```

rm3l (commented):
Added a few comments.

thepetk (approved):
/lgtm

LIFECYCLE.md
```diff
@@ -0,0 +1,55 @@
...
+
+| Role             | Responsibilities                                                                                                                                                                                                                                                                                                           |
+|:-----------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `Stack Provider` | <ul>Follow the [contributing guide](CONTRIBUTING.md) and ensure stack is tested against the supported clients </ul>                                                                                                                                                                                                        |
```
michael-valdron, line 21:
```suggestion
| `Stack Provider` | <ul>Follow the [contributing guide](CONTRIBUTING.md) and ensure stack or sample is tested against the supported clients </ul>                                                                                                                                                                                                        |
```

LIFECYCLE.md
```diff
@@ -0,0 +1,55 @@
...
+| `Devfile Team`   | <ul><li>Review the PR</li><li>Work with `Tooling Clients` to ensure there is sufficient testing</li><li>Ensure the stack has the right contact information and any 2.2 and above versions contain the support link</li><li>Responsible for giving the stack owner repo permissions to modify the stacks they own</li></ul> |
+
+
+## Maintenance
+During the course of its lifecycle, a stack or sample may need to be updated by the `Stack Provider`.  Types of updates can include:
+
+* Changing the devfile content itself where components, commands, resources are modified
+* Using a more secure container image(s)
+* Implementing a new devfile spec version of the existing devfile e.g. two devfiles exist for the same runtime supporting both inner and outer loop scenarios.
```
michael-valdron, line 30: I'm not sure the example given clearly match this scenario and sounds more like a case of having multiple devfiles under the same workspace. We could include this as an update scenario of its own. In addition, I think an example for spec version would be:
> e.g. multiple versions of devfiles exist for the same runtime supporting different versions of devtools

An additional scenario for the inner and outerloop scenarios could go like:
> Implementing new additional devfile to the current stack or sample e.g. two devfiles, `devfile.yaml` and `prod.devfile.yaml`, exist for the same runtime supporting both inner and outer loop scenarios.

LIFECYCLE.md
```diff
@@ -0,0 +1,55 @@
...
+| `Devfile Team`   | <ul><li>Conduct reviews of devfiles in the registries to ensure updates are happening on a regular basis</li><li>Communicate to `Stack Providers` any actions that need to be taken</li></ul>                                                 |
+| `End User`       | <ul>[Report any issues](https://github.com/devfile/registry#reporting-any-issue) with the stacks and samples</ul>                                                                                                                             |
+
+
+## Deprecation
+
+When a stack or sample is no longer maintained due to inactivity, lack of timely updates, dependency on end-of-life (EOL) base images, etc it will be marked deprecated. The `Devfile Team` will reach out to the `Stack Provider` and get agreement before proceeding, but if there is no response within a 3-month timeframe, the  `Devfile Team` will take action and mark the devfile deprecated.
+
+* Deprecated devfiles will remain in the community registry for 1 year before it’s removed.
+* Deprecated devfiles will remain in the product registry indefinitely.
```
michael-valdron, line 46:
```suggestion
```

LIFECYCLE.md
```diff
@@ -0,0 +1,55 @@
...
+| `Devfile Team`   | <ul><li>Conduct reviews of devfiles in the registries to ensure updates are happening on a regular basis</li><li>Communicate to `Stack Providers` any actions that need to be taken</li></ul>                                                 |
+| `End User`       | <ul>[Report any issues](https://github.com/devfile/registry#reporting-any-issue) with the stacks and samples</ul>                                                                                                                             |
+
+
+## Deprecation
+
+When a stack or sample is no longer maintained due to inactivity, lack of timely updates, dependency on end-of-life (EOL) base images, etc it will be marked deprecated. The `Devfile Team` will reach out to the `Stack Provider` and get agreement before proceeding, but if there is no response within a 3-month timeframe, the  `Devfile Team` will take action and mark the devfile deprecated.
+
+* Deprecated devfiles will remain in the community registry for 1 year before it’s removed.
```
michael-valdron, line 45: I wonder if it worth having an additional step that we take to offboard the previous owner and onboard a new owner before removing stack or sample completely (in the case of the entire stack or sample being deprecated).

If no new owner steps up to takeover we could then proceed with deprecating/removing.

michael-valdron (commented):
In addition to what I said in one of my review comment, I think we should include a section for *ownership transfers* in the event a current stack/sample owner can no longer take on this role.

LIFECYCLE.md
thepetk: I think the one year period we have set before removing the completely the stack is sufficient to attract new owners in case someone wants to keep this stack active. So I would say that IMO for now taking an extra step might not be required.

LIFECYCLE.md
thepetk: Yup +1 from me. I think having two different example might make things clearer. I've added your suggestions to the doc. I'm resolving the comment, if you think it needs more refinement we can always re-open!

LIFECYCLE.md
thepetk: @michael-valdron forgot to mention here, that while I don't think is necessary to add an extra step here, I find a good idea to add a section specifically for ownership transfers as you mention here https://github.com/devfile/registry/pull/208#pullrequestreview-1875512655

thepetk: @michael-valdron tried to add a sub-section which explicitly mentions ownership transfers :)

thepetk:
In order to spread the word for the stack's lifecycle PR and get as many reviews as possible for the document, I'm listing here all stack providers. Feel free to review the PR so we can optimize our lifecycle processes:

* @kadel 
* @ajm01
* @maxandersen 
* @ehsavoie 
* @BethGriggs 
* @ibuziuk 
* @l0rd

LIFECYCLE.md
```diff
@@ -0,0 +1,57 @@
...
+| `Devfile Team`    | <ul><li>Communicate with the `Stack Providers` to advise them of the impending deprecation notice if devfile is deemed inactive </li> <li>Label the stack or sample as deprecated</li></ul> |
+| `Stack Provider`  | <ul>Communicate with the `Devfile Team` and agree to have the deprecation notice set up</ul>                                                                                                  |
+| `End User`| <ul>Take note of the deprecation notice and do not use the devfile for production</ul>                                                                                                                                                                                   |
+
+### Ownership Transfers
+
+When a stack provider is no longer able to maintain their stacks, an ownership transfer should take place. The stack provider is responsible to find their replacement. As it is described in the deprecation rules, if a stack is inactive for a long time - because of a failed ownership transfer - it will be marked deprecated.
```
michael-valdron, line 57: What if the previous owner fails to find a replacement in time but someone sees the deprecated stack and wants to pick it up? Would there be a process for someone to take ownership on their own accord?

LIFECYCLE.md
michael-valdron: Left another review for it: https://github.com/devfile/registry/pull/208#discussion_r1492591534

LIFECYCLE.md
thepetk: > What if the previous owner fails to find a replacement in time but someone sees the deprecated stack and wants to pick it up? Would there be a process for someone to take ownership on their own accord?

@michael-valdron I think they should open an issue in order to adopt the stack. I've added a small section for this too!

LIFECYCLE.md
thepetk: Resolving this one! Let's continue in the link you've shared

LIFECYCLE.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+| Role                    | Description                                                                                                                             |
+|:------------------------|:----------------------------------------------------------------------------------------------------------------------------------------|
+| `Devfile Team`          | The team that is responsible for onboarding vendor devfiles for the community registry.                              |
+| `Stack/Sample Provider` | The person or team that is responsible for developing the devfile stack or sample that is available for sharing on our devfile registry. |
+| `Tooling Clients`       | These are the tools that support building devfile-based applications                                                                    |
```
l0rd, line 11: If Eclipse Che and odo are among such clients, then the description is not correct. You don't build devfile based application with Che or odo.
```suggestion
| `Tooling Clients`       | These are the tools that use devfiles                                                                    |
```

PR #230, opened 2023-10-14
Copy additional stack resources inside tmpDir before validating stacks with odov3

thepetk:
### What does this PR do?:

This PR tries to copy all additional resources found inside a stack after `odo init --devfile-path`. This way we are able to use `odo init` in an empty directory and run `odo dev` with all the resources required.

fixes https://github.com/devfile/api/issues/1288

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [x] Documentation
- [x] Check Tools Provider

### DIFF
--- a/tests/odov3/odo_test.go
+++ b/tests/odov3/odo_test.go
@@ -62,8 +62,8 @@ func TestOdo(t *testing.T) {
 	}
 
 	for _, dir := range dirs {
-		path := filepath.Join(stacksPath, dir, "devfile.yaml")
-
+		base := filepath.Join(stacksPath, dir)
+		path := filepath.Join(base, "devfile.yaml")
 		parserArgs := parser.ParserArgs{
 			Path: path,
 		}
@@ -76,7 +76,7 @@ func TestOdo(t *testing.T) {
 		starterProjects, err := devfile.Data.GetStarterProjects(common.DevfileOptions{})
 		g.Expect(err).To(BeNil())
 
-		stack := Stack{name: name, version: version, path: path, starterProjects: starterProjects}
+		stack := Stack{name: name, version: version, path: path, starterProjects: starterProjects, base: base}
 
 		stacks = append(stacks, stack)
 	}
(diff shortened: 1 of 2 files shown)


tests/odov3/odo_test.go
```diff
@@ -158,7 +158,7 @@ var _ = Describe("test starter projects from devfile stacks", func() {
 				// (i.e., without implicitly executing any build and/or run commands), since there won't be any source code.
 				// So here, we just want to make sure that odo dev starts properly (to cover cases for example where the dev container does not end up running).
 				cmpName := fmt.Sprintf("cmp-%s-%s", stack.name, randomString(3))
-				_, _, err := runOdo("init", "--devfile-path", stack.path, "--name", cmpName)
+				_, _, err := runOdo("init", "--devfile", stack.name, "--devfile-version", stack.version, "--name", cmpName)
```
rm3l, line 161: Using `odo init --devfile` will fetch the Stack from the registry. As such, it means that if we have a PR changing a Devfile Stack, we would be testing the (already published) Devfile stack, not the one that was changed in the PR. This, IMO, defeats the purpose of this test spec.
There is an open issue in odo about copying the stack dependencies when running `odo init --devfile-path`: https://github.com/redhat-developer/odo/issues/5713
I can try to see if it can be implemented soon, but meanwhile, I think perhaps another approach could be to copy everything from `stack.path` (except `devfile.yaml`) to the current directory, right after running `odo init --devfile-path`. WDYT?

rm3l (changes requested):

tests/odov3/odo_test.go
thepetk: Oh yes you are right :smiling_face_with_tear: 
My first approach was this one, to try to copy everything from the stack dir. Let me find the previous version of my code and push it again.

thepetk: @rm3l I've pushed the new workaround

rm3l (approved):
Thanks for working on this!

PR #278, opened 2024-01-14
Revert adding all potential filenames inside renovate ignorePaths

thepetk:
Singed-off-by: thepetk <thepetk@gmail.com>

### What does this PR do?:

This PR reverts the addition of the `.devfile.yml`, `devfile.yaml` & `devfile.yml` names into the `ignoredPaths` config variable of the renovate.json file. At the moment the renovate bot autocloses all opened PRs against `devfile.yaml` files. As a result, in order to keep active the automation (until a proper solution is found) we should remove the `devfile.yaml` names from the variable.

The most suitable solution would be to add the `.devfile.yaml` (e.g. `devfile/.devfile.yaml`).

N/A

### PR acceptance criteria:

- [ ] Contributing guide
- [ ] Test automation
- [ ] Documentation
- [ ] Check Tools Provider


renovate.json
```diff
@@ -31,9 +31,6 @@
   "ignorePaths": [
     "**/docker/**",
     ".devfile.yaml",
-    ".devfile.yml",
-    "devfile.yaml",
-    "devfile.yml",
```
michael-valdron, line 36: Since all of these could be used in either case we should fix them to only exclude any of these at the root level rather than any pattern that matches these.

@thepetk Would the following changes work?
```diff
-    ".devfile.yaml",
-    ".devfile.yml",
-    "devfile.yaml",
-    "devfile.yml",
+    "/.devfile.yaml",
+    "/.devfile.yml",
+    "/devfile.yaml",
+    "/devfile.yml",
```

michael-valdron (changes requested):

renovate.json
thepetk: Yeap I think this is a better alternative than creating a separate dir :) @michael-valdron

thepetk: done!

michael-valdron:
@thepetk The fix did not work :thinking: 

- https://github.com/devfile/registry/pull/277
- https://github.com/devfile/registry/pull/275
- https://github.com/devfile/registry/pull/274
- https://github.com/devfile/registry/pull/273

I suspect this might be a bug with renovate if the docs are up to date, should work like `.gitignore`: https://docs.renovatebot.com/configuration-options/#ignorepaths

I'll investigate if there is a bug report for this and open one if there is not.

We could revert the additional devfile ignores to not exclude `devfile.yaml` entries but I strongly feel in the long term this should be fixed if we are to support proper devfile file paths.

michael-valdron:
> I'll investigate if there is a bug report for this and open one if there is not.

**Update**: No bug reported for this, the GitHub issue tracker for renovate does not let the community to report bugs though.

thepetk:
@michael-valdron I suspect the PRs might have been flagged or something. I believe the current config is ok, so Id wait the next run and see how renovate behaves.

PR #342, opened 2024-03-25
Add a new devfile version with an https endpoint

vinokurig:
### What does this PR do?:
Add an https endpoint to be able to run the application on Eclipse CHE 
### Which issue(s) this PR fixes:
https://github.com/eclipse/che/issues/22869

### PR acceptance criteria:

- [ ] Contributing guide
_Have you read the [devfile registry contributing guide](https://github.com/devfile/registry/blob/main/CONTRIBUTING.md) and followed its instructions?_
- [ ] Test automation
_Does this repository's tests pass with your changes?_
- [ ] Documentation
_Does any documentation need to be updated with your changes?_
- [ ] Check Tools Provider
_Have you tested the changes with existing tools, i.e. Odo, Che, Console? (See [devfile registry contributing guide](https://github.com/devfile/registry/blob/main/CONTRIBUTING.md) on how to test changes)_

### How to test changes / Special notes to the reviewer:

PR #345, opened 2024-03-26
Update default version for java-wildfly-bootable-jar stack; set https protocol

olexii4:
### What does this PR do?:
Updates **java-wildfly-bootable-jar** stack:

Add new version 1.3.0
 - Set **HTTPS** protocol for the endpoint
 - Set the **1.3.0** version as the default one

Знімок екрана 2024-03-26 о 17 41 39

Знімок екрана 2024-03-26 о 17 44 02

https://github.com/eclipse/che/issues/22869

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [ ] Documentation
- [x] Check Tools Provider

### DIFF
--- a/stacks/java-wildfly-bootable-jar/1.3.0/devfile.yaml
+++ b/stacks/java-wildfly-bootable-jar/1.3.0/devfile.yaml
@@ -1,0 +1,219 @@
+schemaVersion: 2.2.2
+metadata:
+  name: java-wildfly-bootable-jar
+  displayName: WildFly Bootable Jar
+  description: Java application using WildFly in bootable Jar mode, OpenJDK 11 and Maven 3.6
+  icon: https://design.jboss.org/wildfly/logo/final/wildfly_logomark.svg
+  tags:
+    - RHEL8
+    - Java
+    - OpenJDK
+    - Maven
+    - WildFly
+    - Microprofile
+    - WildFly Bootable
+  projectType: WildFly
+  language: Java
+  version: 1.3.0
+  website: https://docs.wildfly.org/bootablejar/
+starterProjects:
+  - name: microprofile-config
+    description: WildFly Eclipse Microprofile Config Quickstart
+    git:
+      checkoutFrom:
+        remote: wildfly-quickstart
+        revision: 26.1.0.Final
+      remotes:
+        wildfly-quickstart: https://github.com/wildfly/quickstart.git
+    subDir: microprofile-config
+  - name: microprofile-fault-tolerance
+    description: WildFly Eclipse Microprofile Fault Tolerance Quickstart
+    git:
+      checkoutFrom:
+        remote: wildfly-quickstart
+        revision: 26.1.0.Final
+      remotes:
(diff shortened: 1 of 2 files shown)


thepetk (commented):
I'd suggest bumping up to `1.3.0` instead of `1.2.1` because the `schemaVersion` goes from `2.1.0` to `2.2.2`.

@ehsavoie as you are the owner of the stack could you please review the PR?

Another comment is if we would like to remove the version `1.2.0` completely (dir and stack.yaml record) like we did for https://github.com/devfile/registry/pull/342

olexii4:
@thepetk I have fixed the version.

Jdubrick:
> I'd suggest bumping up to `1.3.0` instead of `1.2.1` because the `schemaVersion` goes from `2.1.0` to `2.2.2`.
> 
> @ehsavoie as you are the owner of the stack could you please review the PR?
> 
> Another comment is if we would like to remove the version `1.2.0` completely (dir and stack.yaml record) like we did for #342
> 
> cc @michael-valdron @Jdubrick

I think in this case we leave `1.2.0` because it is unchanged and we are adding `1.3.0` in addition, so people can choose to still reference `1.2.0` if they want by leaving it, wdyt?

Also @olexii4 you will need to sign your latest commit for DCO check to pass :)

thepetk:
> > I'd suggest bumping up to `1.3.0` instead of `1.2.1` because the `schemaVersion` goes from `2.1.0` to `2.2.2`.
> > @ehsavoie as you are the owner of the stack could you please review the PR?
> > Another comment is if we would like to remove the version `1.2.0` completely (dir and stack.yaml record) like we did for #342
> > cc @michael-valdron @Jdubrick
> 
> I think in this case we leave `1.2.0` because it is unchanged and we are adding `1.3.0` in addition, so people can choose to still reference `1.2.0` if they want by leaving it, wdyt?
> 
> Also @olexii4 you will need to sign your latest commit for DCO check to pass :)

Yeah I agree, that's why I had the doubt on this - cause the two versions have different schema versions. +1 on keeping the old version.

olexii4:
> > I'd suggest bumping up to `1.3.0` instead of `1.2.1` because the `schemaVersion` goes from `2.1.0` to `2.2.2`.
> > @ehsavoie as you are the owner of the stack could you please review the PR?
> > Another comment is if we would like to remove the version `1.2.0` completely (dir and stack.yaml record) like we did for #342
> > cc @michael-valdron @Jdubrick
> 
> I think in this case we leave `1.2.0` because it is unchanged and we are adding `1.3.0` in addition, so people can choose to still reference `1.2.0` if they want by leaving it, wdyt?
> 
> Also @olexii4 you will need to sign your latest commit for DCO check to pass :)

I have fixed it.

michael-valdron:
> Another comment is if we would like to remove the version 1.2.0 completely (dir and stack.yaml record) like we did for https://github.com/devfile/registry/pull/342

@thepetk https://github.com/devfile/registry/pull/347#issuecomment-2035372734

ehsavoie (approved):

PR #365, opened 2024-04-12
Fix Nightly Run Failures

Jdubrick:
### What does this PR do?:
This PR fixes the nightly run failures occurring for the odo v2 and odo v3 tests. A summary of changes made are:
- Increased resources allocated to the Minikube instance to `14gb` and `4 cpus`, the allocated resources by default were far too low and could cause issues for some stacks.
- Excluded `java-quarkus` stack from testing as it is not running properly, even in local testing with `odo v2` and `odo v3` it was hanging and never finishing. An investigation issue will need to be opened for this stack.
- Increased `maxTries` from 20 to 50. This is used as a timer while ports are being forwarded, I noticed some java stacks were being timed out at 20 but needed anywhere from 24-50 tries to properly forward.
- Implemented https://github.com/jlumbroso/free-disk-space GitHub action for both the `odov2` and `odov3` jobs. This action frees up a tremendous amount of disk space on the runner which prevents any memory issues while running the tests. This action removes un-required tools from the GitHub runner as they come standard with them.
- Set up a workflow that has `odov2`, `non-terminating` and `validate-devfile-schema` running before `odov3` as `odov3` needs as much available resources as possible. By making it wait for the other jobs it ensures it has access to all the resources.
_Link to github issue(s)_
fixes https://github.com/devfile/api/issues/1480
### PR acceptance criteria:

- [ ] Contributing guide
- [ ] Test automation
- [ ] Documentation
- [ ] Check Tools Provider

GitHub Actions `validate-stacks` should be passing, in my fork it was taking ~1.25-1.50 hours to complete fully.

### DIFF
--- a/.github/workflows/validate-stacks.yaml
+++ b/.github/workflows/validate-stacks.yaml
@@ -131,6 +131,7 @@ jobs:
     name: with odo v3
     runs-on: ubuntu-latest
     needs: [odov2, non-terminating, validate-devfile-schema]
+    if: success() || failure()
     steps:
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
--- a/.github/workflows/validate-stacks.yaml
+++ b/.github/workflows/validate-stacks.yaml
@@ -93,7 +93,7 @@ jobs:
           fetch-depth: 0
       
       - name: Free Disk Space
-        uses: jlumbroso/free-disk-space@main
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be #v1.3.1
         with:
           tool-cache: false
           android: true
@@ -138,7 +138,7 @@ jobs:
           fetch-depth: 0
 
       - name: Free Disk Space
-        uses: jlumbroso/free-disk-space@main
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be #v1.3.1
         with:
           tool-cache: false
           android: true
--- a/.github/workflows/validate-stacks.yaml
+++ b/.github/workflows/validate-stacks.yaml
@@ -29,6 +29,7 @@ concurrency:
 
 env:
   MINIKUBE_VERSION: "v1.29.0"
+  MINIKUBE_RESOURCES: "--memory 14gb --cpus 4"
   KUBERNETES_VERSION: "v1.25.2"
   TEST_DELTA: false
 
@@ -73,7 +74,7 @@ jobs:
           kubernetes version: ${{ env.KUBERNETES_VERSION }}
           driver: "docker"
           github token: ${{ secrets.GITHUB_TOKEN }}
-          start args: "--addons=ingress"
+          start args: "--addons=ingress ${{ env.MINIKUBE_RESOURCES }}"
 
       - name: Test delta if on a pull request
         if: ${{ github.event_name == 'pull_request' }}
--- a/.github/workflows/validate-stacks.yaml
+++ b/.github/workflows/validate-stacks.yaml
@@ -91,6 +91,17 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           fetch-depth: 0
+      
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: true
+          docker-images: true
+          swap-storage: true
 
       - name: Setup Minikube
         uses: manusa/actions-setup-minikube@e2f60d2f9e42413a916781a805284e79ccabaf54 # v2.10.0
@@ -99,7 +110,7 @@ jobs:
           kubernetes version: ${{ env.KUBERNETES_VERSION }}
           driver: "docker"
           github token: ${{ secrets.GITHUB_TOKEN }}
-          start args: "--addons=ingress"
+          start args: "--addons=ingress ${{ env.MINIKUBE_RESOURCES }}"
 
       - name: Install odo v2
         uses: redhat-actions/openshift-tools-installer@2de9a80cf012ad0601021515481d433b91ef8fd5 # v1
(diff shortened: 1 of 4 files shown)


Jdubrick:
I think the tests finished quickly because it's on a PR and it may only test items that were changed. You can see the passing run on my fork here: https://github.com/Jdubrick/registry/actions/runs/8665259912

tests/check_odov2.sh
```diff
@@ -206,7 +206,7 @@ for stack in $STACKS; do
 
   # Skipping the java-wildfly-bootable-jar stack right now since it's broken.
   # TODO: Uncomment once fixed.
-  if [ $stack != "java-wildfly-bootable-jar" ]; then
+  if [[ $stack != "java-wildfly-bootable-jar" && $stack != "java-quarkus/1.3.0" && $stack != "java-quarkus/1.4.0" ]]; then
```
michael-valdron, line 209: `java-quarkus v1.4.0` is not tested by odov2 with or without this statement because it uses a 2.2.x devfile schema, though I'm also good leaving it in anyway since it appears the entire stack is failing the checks right now.

.github/workflows/validate-stacks.yaml
```diff
@@ -118,20 +130,32 @@ jobs:
   odov3:
     name: with odo v3
     runs-on: ubuntu-latest
+    needs: [odov2, non-terminating, validate-devfile-schema]
```
michael-valdron, line 133: Not sure if we should depend on odov2 tests passing in order to run odov3 tests.

I agree with non-terminating and devfile schema validation, though if we do I think this should also be the case for odov2 tests.

.github/workflows/validate-stacks.yaml
```diff
@@ -90,6 +91,17 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           fetch-depth: 0
+      
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@main
```
michael-valdron, line 96: Would it be possible to pin this to a revision hash? Also, use a revision hash of the latest version `v1.3.1` if it has what we need?

.github/workflows/validate-stacks.yaml
```diff
@@ -118,20 +130,32 @@ jobs:
...
     runs-on: ubuntu-latest
+    needs: [odov2, non-terminating, validate-devfile-schema]
     steps:
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           fetch-depth: 0
 
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@main
```
michael-valdron, line 141: > Would it be possible to pin this to a revision hash? Also, use a revision hash of the latest version `v1.3.1` if it has what we need?

Same here.

michael-valdron (changes requested):

tests/check_odov2.sh
Jdubrick: Ah yes that is true. I just added that there so it was consistently being skipped in both `odov2` and `odov3`. I feel that having it setup like that makes it easy to ensure nothing weird happens in the tests, but if you want to remove it I can do that

.github/workflows/validate-stacks.yaml
Jdubrick: I think I can have a workaround that uses conditional statements to have `odov3` run after `odov2` even if `odov2` fails, but I need to look into it because the naive way doesn't let you cancel the workflow if it hangs. The current setup of `odov3` depending on `odov2` is because it needs all the resources it can get. I'll look into how I can skirt around it

Jdubrick: My latest commit that adds `if: success() || failure()` will have `odov3` run even if `odov2` fails but will make sure it always runs AFTER `odov2` so it can have the resources.

tests/check_odov2.sh
michael-valdron: > I feel that having it setup like that makes it easy to ensure nothing weird happens in the tests

:+1:

.github/workflows/validate-stacks.yaml
michael-valdron: > make sure it always runs AFTER `odov2` so it can have the resources.

Just for note keeping, this will ensure _all_ workflow jobs defined under `needs` runs first but will allow `odov3` to run regardless if any of the others fail or not.

tests/odov3/odo_test.go
```diff
@@ -315,7 +315,7 @@ func waitForHttp(url string, expectedCode int) error {
 func waitForPort(devError chan error) ([]ForwardedPort, error) {
 	args := []string{"describe", "component", "-o", "json"}
 
-	maxTries := 20
+	maxTries := 50
```
thepetk, line 318: I'm wondering here, if we also need to increase the timeout too. If we have such cases that need more than 20 retries, does it make sense to increase it to `3h` instead of `2h` (in the `check_odo3.sh)

thepetk: One other thought I had, is maybe we can keep the same amount of retries but increase the delay?

E.g. first retry 10 secs the second will be bigger and so on (keeping always the overall time of retries to the 50*10 secs time so we don't wait for more time)

thepetk (commented):
left some minor comments

tests/odov3/odo_test.go
Jdubrick: I don't mind increasing the timeout to `3h` instead of `2h`, however, I did notice in testing that when a stack does timeout (Quarkus for example and that's why it is skipped) it will timeout every single time and all the others will run fine and relatively quickly. I think on average it was between 1h and 1.5h for the `odov3` testing consistently.

For your other point of increasing the delay we can try that, right now I believe every time it runs the delay increases by `10 seconds` from the previous so it would take some testing to see what the delay needs to be to avoid any issues.

In my opinion I like the idea of increasing the timeout from `3h` from `2h` more because the `maxTries` being set to `50` is already producing the results we're after. Increasing the timeout as you said also will give us benefit of having a buffer incase the GitHub runner is being slow or something and takes more time.

thepetk (approved):
/lgtm

Jdubrick:
FYI I increased the timeout from `2h` to `3h` as discussed above. 

cc @thepetk @michael-valdron

michael-valdron (approved):
/lgtm

PR #357, opened 2024-04-10
chore: Update go devfile

tolusha:
### What does this PR do?:
_Summarize the changes. Are any stacks or samples added or updated?_

* Update schemaversion to 2.2.2
* Update endpoint protocol to https

### Which issue(s) this PR fixes:
_Link to github issue(s)_

### PR acceptance criteria:

- [ ] Contributing guide
_Have you read the [devfile registry contributing guide](https://github.com/devfile/registry/blob/main/CONTRIBUTING.md) and followed its instructions?_
- [ ] Test automation
_Does this repository's tests pass with your changes?_
- [ ] Documentation
_Does any documentation need to be updated with your changes?_
- [ ] Check Tools Provider
_Have you tested the changes with existing tools, i.e. Odo, Che, Console? (See [devfile registry contributing guide](https://github.com/devfile/registry/blob/main/CONTRIBUTING.md) on how to test changes)_

### How to test changes / Special notes to the reviewer:

PR #386, opened 2024-05-08
Adding go 1.21 devfile

ibuziuk:
### What does this PR do?:
Adding go 1.21 devfile

### Which issue(s) this PR fixes:\
N/A

### PR acceptance criteria:

- [x] Contributing guide
- [x] Test automation
- [ ] Documentation
- [x] Check Tools Provider

https://workspaces.openshift.com#https://raw.githubusercontent.com/devfile/registry/747c75aeefb71de0f154c77cf307940a9b7069c1/stacks/go/2.4.0/devfile.yaml

### DIFF
--- a/stacks/go/2.4.0/devfile.yaml
+++ b/stacks/go/2.4.0/devfile.yaml
@@ -1,0 +1,104 @@
+schemaVersion: 2.2.2
+metadata:
+  name: go
+  displayName: Go Runtime
+  description: Go (version 1.21.x) is an open source programming language that makes it easy to build simple, reliable, and efficient software.
+  icon: https://raw.githubusercontent.com/devfile-samples/devfile-stack-icons/main/golang.svg
+  tags:
+    - Go
+  projectType: Go
+  language: Go
+  provider: Red Hat
+  version: 2.4.0
+starterProjects:
+  - name: go-starter
+    description: A Go project with a simple HTTP server
+    git:
+      checkoutFrom:
+        revision: main
+      remotes:
+        origin: https://github.com/devfile-samples/devfile-stack-go.git
+components:
+  - name: build
+    image:
+      imageName: go-image:latest
+      dockerfile:
+        uri: docker/Dockerfile
+        buildContext: .
+        rootRequired: false
+  - name: deploy
+    kubernetes:
+      uri: kubernetes/deploy.yaml
+      endpoints:
+      - name: http-8081
+        targetPort: 8081
+  - container:
+      endpoints:
+        - name: https-go
+          protocol: https
+          targetPort: 8080
+        - exposure: none
+          name: debug
+          targetPort: 5858
+      image: registry.access.redhat.com/ubi9/go-toolset:1.21.9-1.1714671022
+      args: ['tail', '-f', '/dev/null']
+      env:
+        - name: DEBUG_PORT
+          value: '5858'
+      memoryLimit: 1024Mi
+      mountSources: true
+    name: runtime
+commands:
+  - id: build-image
+    apply:
+      component: build
+  - id: deployk8s
+    apply:
+      component: deploy
+  - id: deploy
+    composite:
+      commands:
+        - build-image
+        - deployk8s
+      group:
+        kind: deploy
+        isDefault: true
+  - exec:
+      env:
+        - name: GOPATH
+          value: ${PROJECT_SOURCE}/.go
+        - name: GOCACHE
+          value: ${PROJECT_SOURCE}/.cache
+      commandLine: go build main.go
+      component: runtime
+      group:
+        isDefault: true
+        kind: build
+      workingDir: ${PROJECT_SOURCE}
+    id: build
+  - exec:
+      commandLine: ./main
+      component: runtime
+      group:
+        isDefault: true
+        kind: run
+      workingDir: ${PROJECT_SOURCE}
+    id: run
+
+  - exec:
+      commandLine: |
+        GOPATH=${PROJECT_SOURCE}/.go \
+        GOCACHE=${PROJECT_SOURCE}/.cache \
+        dlv \
+          --listen=127.0.0.1:${DEBUG_PORT} \
+          --only-same-user=false \
+          --headless=true \
+          --api-version=2 \
+          --accept-multiclient \
+          debug --continue main.go
+      component: runtime
+      group:
+        isDefault: true
+        kind: debug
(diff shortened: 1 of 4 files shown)


thepetk (commented):
@ibuziuk in general looks good to me, one minor suggestion is to add also a `1.4.0` version using `go 1.21` & have an `innerloop` approach.

ibuziuk:
@thepetk thanks for the review, but I'm not a fan of the versioning approach in general. Could you clarify what is the difference between go `1.x.x` and `2.x.x` and why do we need both

thepetk:
> @thepetk thanks for the review, but I'm not a fan of the versioning approach in general. Could you clarify what is the difference between go `1.x.x` and `2.x.x` and why do we need both

So far the `1.x.x` versions support `innerloop` scenarios and the `2.x.x` `outerloop` scenarios.

ibuziuk:
@thepetk any reason why not to have one devfile that combines both scenarios?

thepetk:
> @thepetk any reason why not to have one devfile that combines both scenarios?

I think we already combine both scenarios. From my perspective, your current approach of the `2.4.0` already covers both approaches (like most of the 2.x.x approaches). My point for "backporting" the update to go 1.21 also in `1.x.x` versions, is that the current default version of the `go` stack is `1.x.x`. So IMO creating the `1.4.0` stack version could be good addition.

ibuziuk:
@thepetk @michael-valdron folks, I updated the PR based on your feedback, but I do not really understand what you are trying to achieve with versioning. Yes, I do understand that sometimes there could be compatibility reasons, but this does not seem to be the case. Why do you want to continue having 1.x / 2.x versions pattern of Go devfiles. For me personally it is super confusing. I would even understand if you would like to have identical devfiles with only go version diff e.g. 1.19, 1.20, 1.21 etc. but this inner / outer loop approach with 1.x  / 2.x does not make any sense imo

stacks/go/1.4.0/devfile.yaml
```diff
@@ -0,0 +1,79 @@
+schemaVersion: 2.2.2
```
michael-valdron, line 1: 1.x stack versions should be still using devfile 2.1.0:
```suggestion
schemaVersion: 2.1.0
```

michael-valdron (changes requested):
In addition to my comment, I'd also recommend deprecating the patch versions introduced by https://github.com/devfile/registry/pull/357 to prompt anyone using these versions to either bump to `1.4.0` or switch to `2.x` versions if they need to use the `2.2.x` devfile specification.

michael-valdron:
> @thepetk @michael-valdron folks, I updated the PR based on your feedback, but I do not really understand what you are trying to achieve with versioning. Yes, I do understand that sometimes there could be compatibility reasons, but this does not seem to be the case. Why do you want to continue having 1.x / 2.x versions pattern of Go devfiles. For me personally it is super confusing. I would even understand if you would like to have identical devfiles with only go version diff e.g. 1.19, 1.20, 1.21 etc. but this inner / outer loop approach with 1.x / 2.x does not make any sense imo

Difference of separating 1.x / 2.x is that some tools don't have compatibility with the 2.2.x devfile spec, such as odo v2, this is why 1.x was left with just an inner loop setup and 2.x includes both. Even if a tool only supports inner loop the tool _should_ still use 2.x version of the stack if the tool supports the 2.2.x devfile spec.

The change merged in https://github.com/devfile/registry/pull/357 however has bumped the schema version for 1.x to 2.2.x devfile spec, an oversight on our part. I would recommend [reverting the 1.x stack back to using devfile 2.1.0](https://github.com/devfile/registry/pull/386#discussion_r1605251805) to maintain the purpose of this stack version and if there are use cases need to use the 2.2.x specification to switch to the 2.x stack instead.

Apologies for the mix up and misunderstanding here.

cc @ibuziuk @thepetk

stacks/go/1.4.0/devfile.yaml
ibuziuk: why?

michael-valdron: https://github.com/devfile/registry/pull/386#issuecomment-2117944019

ibuziuk:
@michael-valdron I removed 1.x version from the PR. Feel free to merge it in the current state or close since I do not plan to provide any more updates

michael-valdron:
Using https://github.com/devfile/api/issues/1564 to track these changes towards the v1 go stack.

> The change merged in https://github.com/devfile/registry/pull/357 however has bumped the schema version for 1.x to 2.2.x devfile spec, an oversight on our part. I would recommend https://github.com/devfile/registry/pull/386#discussion_r1605251805 to maintain the purpose of this stack version and if there are use cases need to use the 2.2.x specification to switch to the 2.x stack instead.

This issue is being by https://github.com/devfile/api/issues/1563 and will be handled by us.

michael-valdron (commented):
@ibuziuk Changes looks good to me, need a sign off from someone on your team then I can sign off on this.

cc @devfile/che-team

stacks/go/2.4.0/devfile.yaml
```diff
@@ -0,0 +1,103 @@
...
+        rootRequired: false
+  - name: deploy
+    kubernetes:
+      uri: kubernetes/deploy.yaml
+      endpoints:
+      - name: http-8081
+        targetPort: 8081
+  - container:
+      endpoints:
+        - name: http-go
```
svor, line 37:
```suggestion
        - name: https-go
          protocol: https
```

svor (commented):
I was able to start a workspace on Che, but commands are failed:
screenshot-devspaces apps sandbox-m4 g2pi p1 openshiftapps com-2024 05 20-10_24_52

I think we have an issue for that: https://github.com/eclipse-che/che/issues/22930

Also if we want to see this version on the Che Dashboard, need to mark it as a default one in stack.yaml file

svor:
@tolusha am I right that it's a known issue that commands from Go sample are failing on Che

tolusha:
@svor 
My guess, this issue [1] is mostly about nonsupporting env vars in the commands. At least it looks like.
As a result, the following exec command fails:
```yaml
 exec:
      env:
        - name: GOPATH
          value: ${PROJECT_SOURCE}/.go
        - name: GOCACHE
          value: ${PROJECT_SOURCE}/.cache
      commandLine: go build main.go
```

[1] https://github.com/eclipse-che/che/issues/22930

svor (approved):
lgtm

PR #397, opened 2024-05-14
[WIP]Add parent dependency check on samples

michael-valdron:
### What does this PR do?:

Adds new script and new workflow to perform testing on child samples of stacks.

fixes https://github.com/devfile/api/issues/1142

### PR acceptance criteria:

- [ ] Contributing guide
- [ ] Test automation
- [ ] Documentation
- [ ] Check Tools Provider


michael-valdron:
> Check Tools Provider

CI changes, N/A

tests/build_parents_file.sh
```diff
@@ -0,0 +1,251 @@
+#!/bin/bash
```
thepetk, line 1: Minor but not sure name-wise if `build_parents_file` includes all the functionality as we also getting all children for the parents.

tests/build_parents_file.sh
```diff
@@ -0,0 +1,251 @@
...
+                $YQ_PATH eval ".parents[${next_idx}].version = \"${parent_version}\"" -i ${parents_file}
+            fi
+        fi
+    else
+        return 1
+    fi
+}
+
+# Builds children of parent stacks
+build_children() {
```
thepetk, line 135: I was thinking if it makes sense to rename this to `build_child` (singular) as (if I'm not wrong) we only build a specific sample version.

tests/build_parents_file.sh
```diff
@@ -0,0 +1,251 @@
...
+
+    if [ "${result}" == "null" ]; then
+        echo "-1"
+    else
+        echo ${result}
+    fi
+}
+
+# Builds sample parents
+build_parents() {
```
thepetk, line 94: Similar to `build_children` does it makes sens to rename this to `build_parent`?

.gitignore
```diff
@@ -3,4 +3,6 @@ registry-support/
 .idea/
 devfile-web/
 vendor/
-.odo
\ No newline at end of file
+.odo
+samples/.cache
```
thepetk, line 7: Shouldn't we add the `samples` dir in general?

tests/validate_devfile_schemas.sh
```diff
@@ -1,8 +1,42 @@
...
+
+# Use pwd if relative path
+if [[ ! ${stacksDir} = /* ]]; then
+  stacksDir=$(pwd)/${stacksDir}
+fi
+
+# Unzip resource files if samples
+if [ "${SAMPLES}" == "true" ]; then
+  for sample_dir in $(ls $stacksDir); do
+    unzip -n $stacksDir/$sample_dir/sampleName.zip -d $stacksDir
```
thepetk, line 37: Here we could add again an option for a more verbose output. I'd suggest including `-q` for the default. If `VERBOSE="true"` then we can run the command without `-q`. WDYT?

thepetk (commented):
In general looks good to me, left only some minor comments. Tested locally and it works. Very nice addition :)

.gitignore
thepetk: Yes +1 for the `.cache/samples` dir sounds better!

Jdubrick (commented):
Overall looks good to me. I pushed your branch to my own repo and tested the workflow and it works great! After changes Fanis suggested are made I think this is good to get approved

tests/validate_devfile_schemas.sh
michael-valdron: Sounds good to me :+1:

tests/build_parents_file.sh
michael-valdron: @thepetk The name is referring to the `parents.yaml` file this script builds rather than referring to everything it does to build the file, though I'm open to suggestions to improve the descriptor.

AObuchow:
Just thinking out loud: the logic introduced in this PR could potentially be adopted to resolved https://github.com/devfile/registry/pull/401#issuecomment-2134023630 -- good timing :)

tests/build_parents_file.sh
thepetk: Yeah makes sense. TBH I don't have any good suggestions atm so let's keep it this way and we can revisit in the future if necessary :)

thepetk (approved):
/lgtm really helpful check!

Jdubrick (approved):

PR #401, opened 2024-05-22
Using parent defile for Ollama

ibuziuk:
### What does this PR do?:
Using parent defile for Ollama devfile

_Link to github issue(s)_

### PR acceptance criteria:

- [ ] Contributing guide
- [ ] Test automation
- [ ] Documentation
- [ ] Check Tools Provider


ibuziuk:
@thepetk @elsony folks, any ideas why we are hitting `Some Attributes are already defined in parent: controller.devfile.io/storage-type. If you want to override them, you should do it in the parent scope.` - https://github.com/devfile/registry/actions/runs/9195275297/job/25290700742#step:4:891 ?

it is not possible to reference parent that is ephemeral ? cc: @AObuchow to follow up during my PTO

thepetk:
> @thepetk @elsony folks, any ideas why we are hitting `Some Attributes are already defined in parent: controller.devfile.io/storage-type. If you want to override them, you should do it in the parent scope.` - https://github.com/devfile/registry/actions/runs/9195275297/job/25290700742#step:4:891 ?
> 
> it is not possible to reference parent that is ephemeral ? cc: @AObuchow to follow up during my PTO

I've taken a thorough look to see if there's any issue with the way we merge the flatten content and we check overrides but I haven't found anything. I would expect to see this error in case both the stack devfile & the parent devfile define the same attribute (e.g. `controller.devfile.io/storage-type: ephemeral`).

More detailed if the parent already defines this attribute the correct way to override it would be:
```yaml
parent:
    attributes:
        controller.devfile.io/storage-type: ephemeral
```

As I see for the current state of the PR [this check is successful](https://github.com/devfile/registry/actions/runs/9195434450/job/25291250098?pr=401) I suspect that the mentioned check is for this [commit](https://github.com/devfile/registry/commit/1d0f7728ffeaf6025d011959532b063f945896d6), only because this is a case that we should raise an error that someone is trying to override an attribute already set by the parent.

AObuchow:
I believe the check for non-terminating images is [failing](https://github.com/devfile/registry/actions/runs/9259636857/job/25472255936?pr=401) because the [logic](https://github.com/devfile/registry/blob/main/tests/check_non_terminating.sh#L37-L43) for getting the first container component's image is not able to fetch container components from the [parent devfile](https://raw.githubusercontent.com/redhat-developer-demos/cde-ollama-continue/main/devfile.yaml). This is why we're seeing `null` for the component image in the [command output](https://github.com/devfile/registry/blob/main/tests/check_non_terminating.sh#L87): 

```
Testing /home/runner/work/registry/registry/stacks/ollama/devfile.yaml
  COMMAND: minikube ssh docker pull null
```

stacks/ollama/devfile.yaml
```diff
@@ -11,55 +11,6 @@ metadata:
...
-      component: udi
-      commandLine: "mkdir /home/user/.continue && cp /projects/cde-ollama-continue/continue-config.json /home/user/.continue/config.json"
-events:
-  postStart:
-    - pullmodel
-    - pullautocompletemodel
-    - copyconfig
+  version: 1.0.1
+parent:
+  uri: https://raw.githubusercontent.com/redhat-developer-demos/cde-ollama-continue/main/devfile.yaml
```
thepetk, line 16: I think here it would be better to use a specific commit instead of the main branch. Wdyt?

AObuchow: This makes sense given that we are versioning the devfile (in this case, 1.0.1). We might want to instead have a branch in the parent devfile repo for each devfile version, and "release" new versions of the devfile, rather than having a rolling release. This is definitely something to gather consensus for https://github.com/devfile/api/issues/1579.

thepetk: Yes +1 I think is important to create some guidelines regarding the versioning of every parent devfile too. This is one of the reasons the spike was opened. For using a release branch I think it also makes sense.

cc @devfile/devfile-services-team in case someone else has input for the usage of a release branch.

thepetk:
@ibuziuk now that the https://github.com/devfile/api/issues/1568 is completed and the `check-non-terminatng` check is flattening the parent devfile, we should rebase this PR to see if everything works ok.

michael-valdron:
Closing this PR as stale.

PR #403, opened 2024-05-23
Update `Node.js Runtime` stack

azatsarynnyy:
### What does this PR do?:
_Summarize the changes. Are any stacks or samples added or updated?_
- update schema to `2.2.0`;
- use `https` protocol for the `nodejs` devfile

### Which issue(s) this PR fixes:
_Link to github issue(s)_
https://github.com/eclipse-che/che/issues/22869

### PR acceptance criteria:

- [ ] Contributing guide
_Have you read the [devfile registry contributing guide](https://github.com/devfile/registry/blob/main/CONTRIBUTING.md) and followed its instructions?_
- [ ] Test automation
_Does this repository's tests pass with your changes?_
- [ ] Documentation
_Does any documentation need to be updated with your changes?_
- [ ] Check Tools Provider
_Have you tested the changes with existing tools, i.e. Odo, Che, Console? (See [devfile registry contributing guide](https://github.com/devfile/registry/blob/main/CONTRIBUTING.md) on how to test changes)_

### How to test changes / Special notes to the reviewer:
https://workspaces.openshift.com/#https://raw.githubusercontent.com/azatsarynnyy/registry/patch-1/stacks/nodejs/2.2.0/devfile.yaml

PR #418, opened 2024-06-06
Update stack.yaml

yangcao77:
change back the default version to stack version `2.2.0`

### What does this PR do?:

_Link to github issue(s)_

### PR acceptance criteria:

- [ ] Contributing guide
- [ ] Test automation
- [ ] Documentation
- [ ] Check Tools Provider

### DIFF
--- a/stacks/nodejs/stack.yaml
+++ b/stacks/nodejs/stack.yaml
@@ -5,5 +5,5 @@ icon: https://raw.githubusercontent.com/devfile-samples/devfile-stack-icons/main
 versions:
   - version: 2.1.1
   - version: 2.2.0
-  - version: 2.2.1
     default: true # should have one and only one default version
+  - version: 2.2.1


michael-valdron:
FYI this is to restore compatibility with tools still using `odo v3.15.0` which does not support `> 2.2.0` schema versions.

michael-valdron:
cc @lholmquist

michael-valdron (commented):
@yangcao77 you'll need to sign-off

thepetk (commented):
I was wondering if the best approach could be here to downgrade the `nodejs/2.2.1` to use `2.2.0` schemaVersion.

More detailed, IIRC, the @devfile/che-team is consuming always the `default` stack version. If we fallback to default nodejs stack version -> `2.2.0`, most probably we will undo on the che-side the work done for https://github.com/devfile/registry/pull/403.

That said, I think might be better to update the current default to use devfile schema version `2.2.0` ([here](https://github.com/devfile/registry/blob/main/stacks/nodejs/2.2.1/devfile.yaml#L1)).

michael-valdron:
> I was wondering if the best approach could be here to downgrade the `nodejs/2.2.1` to use `2.2.0` schemaVersion.
> 
> More detailed, IIRC, the @devfile/che-team is consuming always the `default` stack version. If we fallback to default nodejs stack version -> `2.2.0`, most probably we will undo on the che-side the work done for #403.
> 
> That said, I think might be better to update the current default to use devfile schema version `2.2.0` ([here](https://github.com/devfile/registry/blob/main/stacks/nodejs/2.2.1/devfile.yaml#L1)).
> 
> @michael-valdron @yangcao77 wdyt? (I've also pinged che-team to confirm that they always consume the default)
> 
> /hold

@thepetk I'm good with this approach as long as @devfile/che-team and @lholmquist are also good with this approach.

If we want to preserve a version with `2.2.2` schema version, I can also suggest creating `2.3.0` for the current state and keeping `2.2.1` as the default with the schema version fallback to `2.2.0`.

PR #504, opened 2024-12-04
Add stack break arguments from jhipster-online version 2.33.0 to ginkgo test

maximilianoPizarro:
# Description of Changes

Just add jhipster-online version: 2.33.0 jump stack arguments to ginkgo test because fail in step with odo v3.

For example:
https://github.com/devfile/registry/actions/runs/12121480511/job/33795062809


Jdubrick (commented):
Is it possible to rebase to registry main? I'm not sure why so many commits that are already present in `main` are being added

PR #574, opened 2025-01-31
Flatten stacks with parent in non-terminating-image checks

thepetk:
# Description of Changes
The PR fixes the issue that the `check_non_terminating` script faces for stacks that include a parent. An example is https://github.com/devfile/registry/pull/401 where the changes introduced are failing.

The list of updates:
* We add a build step to the workflow.
* The build step refers to a small golang bin file which utilizes the library to parse, flatten and update the devfile.
* The necessary updates were added in the bash script to check if a parent is included in a devfile and if so to run the bin and flatten it.

# Related Issue(s)

fixes https://github.com/devfile/api/issues/1568

# Acceptance Criteria

- [ ] Contributing guide

- [ ] Test automation

- [ ] Documentation

- [ ] Check Tools Provider

# Tests Performed
_Explain what tests you personally ran to ensure the changes are functioning as expected._

# How To Test

* Update a stack to point to a parent.
* Run minikube or create a cluster
* From project root build the binary:
```bash
go build -C tests/check_non_terminating
```
* Update the STACKS variable inside the tests/check_non_terminating to include only the stack you have temporarily update.
* Finally again from project root, run locally the tests/check_non_terminating.sh script.

# Notes To Reviewer

I've tried to mimic the updates introduced by the PR for ollama and ran the workflow here: https://github.com/thepetk/registry/actions/runs/13076640505/job/36490341149


.github/workflows/validate-stacks.yaml
```diff
@@ -81,7 +81,9 @@ jobs:
         run: echo "TEST_DELTA=true" >> $GITHUB_ENV
 
       - name: Check that containers components are non terminating
-        run: bash tests/check_non_terminating.sh
+        run: |
+          go build -C tests/check_non_terminating
```
Jdubrick, line 85: I'm not familiar with the `-C` flag and I can't find much on it, what does it do?

tests/check_non_terminating.sh
```diff
@@ -7,6 +7,9 @@ DEVFILES_DIR="$(pwd)/stacks"
 # The stacks to test as a string separated by spaces
 STACKS=$(bash "$(pwd)/tests/get_stacks.sh")
 
+# Path to the check_non_terminating go package
+NON_TERMINATING_MODULE_BIN="$(pwd)/tests/check_non_terminating/./check_non_terminating"
```
Jdubrick, line 11:
```suggestion
NON_TERMINATING_MODULE_BIN="$(pwd)/tests/check_non_terminating/check_non_terminating"
```

Do we need the `.` there?

Jdubrick: Also if the second `check_non_terminating` is the binary name maybe we rename it for clarity? So it isn't doubled up in the path

Jdubrick (commented):
Couple comments/questions!

.github/workflows/validate-stacks.yaml
thepetk: I wanted to avoid changing dirs with cd so I used the flag -C to change it on build -> [here](https://pkg.go.dev/cmd/go)

tests/check_non_terminating.sh
thepetk: I made it a bit more dynamic and gave it the name `flatten-parent` as default (also for the workflow).

Jdubrick (approved):
/lgtm

